Issue #24, opened 2025-01-05
Add some kind of indicator for how old data is


motinis:
I don't know how best to display this considering the limited space available.

My thought currently is to rely on coloring of the arrow - yes people will need to read the guide, but I guess it's ok.

E.g.:

White: data < 7 minute old
Light Blue: data < 12 minutes old
Blue: data < 17 minutes old
Yellow: data < 22 minutes old
Orange: data < 27 minutes old
Red: data < 32 minutes old

Then past that, rather than having the BG display in white, to just write OLD DATA in red or orange. People maybe also would want an alarm for this (or an option to configure a stale data alarm after X minutes). Once data is too old, people should be checking other data sources to see what's going on.

Fransysco:
I came here to request something similar. 

I was thinking maybe add a block for counting each minute since last update under the clock face (could color code it as well as it progresses). In my poorly photoshoped images I did 4 pixels for each minute, but 3 or 2 would be enough to see from a distance if more than 6 minutes after update is needed. For me, anything after 5 minutes means dexcom isn't updating and I need to check why.

After a certain point it would probably help to just have "No Data" or something similar displayed.

![clock4](https://github.com/user-attachments/assets/7b7a17e1-a2c6-48a5-a91d-ede5220d3486)
![clock1](https://github.com/user-attachments/assets/5a325051-8c31-4914-a87c-f5ff15411006)
![clock3](https://github.com/user-attachments/assets/d00d3c2f-d994-4224-a9cd-31b77da24512)
![clock2](https://github.com/user-attachments/assets/2e08d4e9-f65d-4e90-b3f2-8b759c448409)

PR #41, opened 2025-04-27
Custom No Data Timer

TheBeardedTechGuy:
These changes add a new field under "Advanced Settings" that allows users to adjust how long it takes for the clock face to turn gray after not receiving data. Default is 20 minutes and can be adjusted from 6 to 60 minutes. Values below 6 would constantly cause the face to be gray. 

It may be better to have this field under "Device Settings".

Please take a good look through this, it worked on my two clocks but it's been a while since I've done any kind of coding :)


TheBeardedTechGuy:
This is to try to help with #24 but I can't figure out how to actually draw the boxes as was suggested.

TheBeardedTechGuy:
> Thanks a lot again! I have merged this PR and added some small changes to be sure that wrong values to not reach the variable. [4eb564f](https://github.com/ktomy/nightscout-clock/commit/4eb564fcbea90be42edc41dc6dbaffbdcb0858df) Please take a look and let me know if you think these changes are ok. Thank

Looks good!

With this and some changes I want to add for the No data timer blocks, I think it might make sense to pull the setting out from advanced settings and add to its own section, but that is for later

PR #43, opened 2025-05-16
Add Timer Blocks

TheBeardedTechGuy:
This will add a timer block function that will add blocks at the bottom of most clock faces based on how many minutes have passed since the last data refresh up to 5 blocks (5 minutes). Blocks turn red after the "No Data" timer expires. This works with (but not required for) https://github.com/ktomy/nightscout-clock/pull/41 that I submitted as well.

Added to clock faces:
Simple
Full Graph
Graph and BG
Value and Diff

Not on:
Big Text - It fits but doesn't look good. Maybe a different color for the timer blocks would make it look better but not sure.
Clock and value - The AM/PM bar is where the timer blocks go

Pictures of clock faces with the timer blocks:
image
image
image
image
image

### DIFF
--- a/src/BGDisplayFaceGraph.cpp
+++ b/src/BGDisplayFaceGraph.cpp
@@ -5,4 +5,15 @@
 
 void BGDisplayFaceGraph::showReadings(const std::list<GlucoseReading> &readings, bool dataIsOld) const {
     showGraph(0, 32, 180, readings);
+
+    auto lastReading = readings.back();
+
+    // Calculate time since last data update
+    int elapsedMinutes = (ServerManager.getUtcEpoch() - lastReading.epoch) / 60;
+
+    // Call timer block function
+    BGDisplayManager_::drawTimerBlocks(elapsedMinutes, 5, dataIsOld);
+
+    DisplayManager.update();
+    
 }
--- a/src/BGDisplayFaceGraphAndBG.cpp
+++ b/src/BGDisplayFaceGraphAndBG.cpp
@@ -15,14 +15,6 @@ void BGDisplayFaceGraphAndBG::showReadings(const std::list<GlucoseReading> &read
 
     auto lastReading = readings.back();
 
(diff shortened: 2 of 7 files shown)


ktomy:
pfff, copilot is too smart for me :) Let me merge the PR, I will keep it enabled for Simple and value&delta faces and than we'll think what to do with the rest of the faces.

ktomy:
@TheBeardedTechGuy please see https://github.com/ktomy/nightscout-clock/commit/13e53aef01d511aa7323a12ad0638e023bea7db2
I tried to improve your code. I made blocks scalable so that they can be placed on more clock faces, including the one with graph and clock.
I have also changed colors. If you don't like that for the old data the blocks become gray, please change colors near https://github.com/ktomy/nightscout-clock/blob/main/src/BGDisplayManager.cpp#L146

Thanks again for your great contribution!

TheBeardedTechGuy:
Great idea with resizing to fit the blocks under the values for graph and time clock faces, it fits well!

I think color is going to be a preference (maybe it can be select-able in the future as a drop down in settings, but the red does draw attention more than gray:
image



I'm not sure if it was intended or not but the timer blocks are a bit dimmer compared to the BG value

(left is your codebase, right is mine):
image

If this wasn't intentional it might be from 1e5ec3a5c72b3d6a79023621bba08c6f148bc825, I also noticed that the auto brightness is not functioning the same anymore which I think is from 1e5ec3a5c72b3d6a79023621bba08c6f148bc825 but was going to open a issue for that.

(left is your codebase, right is mine; both on auto brightness):
image

ktomy:
I have made colors of the blocks darker than digits so that the user draws more attention to the numbers as the timing is an additional information.
For the auto brightness issue, it may happen that the dimming algo has changed in the new version of that library. I will retest and adjust during the day, but maybe it was just a momentary value of the light sensor, I can see that on other faces the brightness of both versions is the same

PR #52, opened 2025-07-06
Add dimmer and brighter brightness profile options

TheBeardedTechGuy:
Adds two new modes to clock that adjusts brightness under "Advanced Settings":

Dimmer - Decreases overall brightness levels to allow for less light pollution in darker areas (such as a bedroom)
Brighter - Increases overall brightness levels to allow for better visibility in brighter environments (such as outside or in front of a window)

I did have to modify some of the coloring for the dimmer brightness profile as at lowest setting the colors weren't displaying. I also didn't want to push the brighter brightness profile too high as I'm not sure what the limitations of the clock are.

I wrote this with being able to maybe trigger these brightness profiles by API calls and I would like to add functionality to be able to pick some of the coloring for the clock in the future.

image
image


ktomy:
Thanks a lot for this change. I will be back home in the end of the week and will take a look. I would integrate it to the main brightness parameter, but I do not know if it is possible without looking at the code

ktomy:
Aha, I have read the code now. Integrating with the brightness regulation would mean changing colors instead of setting brightness on sensor value change, which is simpler than introducing new settings. Anyway, I will take a look and see what is possible. I will also add api endpoint to set brightness.

PR #67, opened 2025-09-04
Updated dimness for dark rooms

unxmaal:
- Sets a non-zero floor (1/255) so display never turns off completely
- Removed the old low-light gain
- Uses a gentle gamma curve so darkness maps to the floor value
- setBrightness() now updates BRIGHTNESS


ktomy:
I added all your code, but made it optional, now users can select between auto brightness modes. See https://github.com/ktomy/nightscout-clock/commit/758863710df2edda0a35a35eb0c27791ecf39cb0

PR #120, opened 2025-12-23
fix: multiple patients using LibreLinkUp always shows first patient

andr3i1010:


ktomy:
Thanks, inspired by this PR I implemented the fix: https://github.com/ktomy/nightscout-clock/commit/855aea9c70787c933bf55346df9bba9027374cef

PR #153, opened 2026-04-01
Add filesystem self-check on boot

miqcie:
## Summary

Follow-up to #143, which was closed with feedback to split into focused PRs. This is the first of 3 focused changes ktomy identified as promising.

- `SettingsManager::setup()` now returns `bool` — validates LittleFS mount and checks that critical files (`index.html`, `config_initial.json`) exist before proceeding
- On failure, displays an actionable error message on the LED matrix: "FS empty - reflash firmware+filesystem via USB"
- `main.cpp` guards the rest of `setup()` and `loop()` with a `setupFailed` flag, preventing crashes from null pointer dereferences when the filesystem is missing
- Removes the `factoryReset()` call in `readConfigJsonFile()` when `config.json` is missing, which could cause an infinite reboot loop on a truly empty filesystem

## Motivation

First-time flashers commonly upload only the firmware binary without the LittleFS filesystem image. The current code proceeds into setup with a missing filesystem, leading to cryptic crashes or infinite reboot loops. This change catches the problem early and shows the user exactly what to do.

## Test plan

- [ ] Flash firmware **without** filesystem image — verify "FS empty" message appears on display and device does not reboot-loop
- [ ] Flash firmware **with** filesystem image — verify normal boot proceeds as before
- [ ] Delete `config.json` from filesystem — verify device does not reboot-loop (no more `factoryReset()` call on missing config)
- [ ] Verify build compiles cleanly with `pio run`

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -12,6 +12,7 @@
 #include "improv_consume.h"
 
 float apModeHintPosition = MATRIX_WIDTH;  // Start the scrolling right after the screen
+bool setupFailed = false;
 
 void setup() {
     pinMode(15, OUTPUT);
@@ -21,9 +22,15 @@ void setup() {
     // Serial.setDebugOutput(true);
 
     DisplayManager.setup();
-    SettingsManager.setup();
+    if (!SettingsManager.setup()) {
+        DisplayManager.showFatalError("FS empty - reflash firmware+filesystem via USB");
+        setupFailed = true;
+        return;
+    }
     if (!SettingsManager.loadSettingsFromFile()) {
         DisplayManager.showFatalError("Error loading software, please reinstall");
+        setupFailed = true;
+        return;
     }
 
     DisplayManager.applySettings();
@@ -74,6 +81,10 @@ void showJoinAP() {
 }
 
 void loop() {
+    if (setupFailed) {
+        delay(1000);
+        return;
+    }
 #ifdef DEBUG_MEMORY
 
     static unsigned long lastMemoryCheck = 0;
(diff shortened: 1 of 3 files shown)


src/main.cpp
```diff
@@ -21,9 +22,15 @@ void setup() {
     // Serial.setDebugOutput(true);
 
     DisplayManager.setup();
-    SettingsManager.setup();
+    if (!SettingsManager.setup()) {
+        DisplayManager.showFatalError("FS empty - reflash firmware+filesystem via USB");
+        setupFailed = true;
```
ktomy, line 27: Execution will never reach this place as `showFatalError` doesn't return, it restarts the clock

src/main.cpp
```diff
@@ -21,9 +22,15 @@ void setup() {
...
+        return;
+    }
     if (!SettingsManager.loadSettingsFromFile()) {
         DisplayManager.showFatalError("Error loading software, please reinstall");
+        setupFailed = true;
```
ktomy, line 32: Execution will never reach this place as showFatalError doesn't return, it restarts the clock

src/main.cpp
```diff
@@ -74,6 +81,10 @@ void showJoinAP() {
 }
 
 void loop() {
+    if (setupFailed) {
```
ktomy, line 84: I don't think it would be a good idea to have an empty loop in case of an error, instead we should do `showFatalError("something")` which shows the error to the user and restarts the clock. In this way we resurrect the clock if something bad happened. Well, if the FS is damaged, clock restarts will not help, but maybe that was a glitch or something.

ktomy (changes requested):
I agree to the additional checks, but I don't agree with the "empty loop", but that is a minor change.

ktomy:
Closing as the author did not apply requested changes

PR #154, opened 2026-04-01
Dexcom credential UX improvements and JSON injection fix

miqcie:
Follow-up to #143 (closed per request for focused PRs). Two changes: Dexcom credential UX improvements and a small security hardening fix.

**Dexcom credential UX (data/index.html)**
- Add autocomplete=username and autocomplete=current-password so password managers can autofill
- Add placeholder text and helper text clarifying these are your Dexcom app credentials

**JSON injection hardening (src/ServerManager.cpp)**
- Add jsonStringEscape() helper for safe string embedding in concatenated JSON
- Apply to bgSource and bgSourceStatus in /api/status endpoint
- Defensive: current values are hardcoded, but prevents future injection risk

No captive portal, weather face, zip geocoding, or deferred restart included.

Builds clean: pio run -e ulanzi (71% flash, 17.6% RAM).

### DIFF
--- a/src/ServerManager.cpp
+++ b/src/ServerManager.cpp
@@ -14,6 +14,33 @@
 #include "globals.h"
 #include "time.h"
 
+// Escape a string for safe embedding in JSON string values.
+// Prevents injection when building JSON responses by concatenation.
+static String jsonStringEscape(const String& input) {
+    String output;
+    output.reserve(input.length());
+    for (unsigned int i = 0; i < input.length(); i++) {
+        char c = input.charAt(i);
+        switch (c) {
+            case '"':  output += "\\\""; break;
+            case '\\': output += "\\\\"; break;
+            case '\n': output += "\\n"; break;
+            case '\r': output += "\\r"; break;
+            case '\t': output += "\\t"; break;
+            default:
+                if (c < 0x20) {
+                    char buf[7];
+                    snprintf(buf, sizeof(buf), "\\u%04x", (unsigned char)c);
+                    output += buf;
+                } else {
+                    output += c;
+                }
+                break;
+        }
+    }
+    return output;
+}
+
 // The getter for the instantiated singleton instance
 ServerManager_& ServerManager_::getInstance() {
     static ServerManager_ instance;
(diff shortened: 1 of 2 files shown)


src/ServerManager.cpp
```diff
@@ -490,9 +517,9 @@ void ServerManager_::setupWebServer(IPAddress ip) {
         jsonResponse += ", \"isInAPMode\": ";
         jsonResponse += this->isInAPMode ? "true" : "false";
         jsonResponse += ", \"bgSource\": \"";
-        jsonResponse += toString(bgSourceManager.getCurrentSourceType());
+        jsonResponse += jsonStringEscape(toString(bgSourceManager.getCurrentSourceType()));
```
ktomy, line 520: This is not a user-entered data, why do you want to sanitize it?

src/ServerManager.cpp
```diff
@@ -490,9 +517,9 @@ void ServerManager_::setupWebServer(IPAddress ip) {
...
         jsonResponse += "\", \"bgSourceStatus\": \"";
-        jsonResponse += bgSourceManager.getSourceStatus();
+        jsonResponse += jsonStringEscape(bgSourceManager.getSourceStatus());
```
ktomy, line 522: statuses are hardcoded values, they don't come from APIs and are not part of the user input, why sanitizing it?

src/ServerManager.cpp
```diff
@@ -14,6 +14,33 @@
 #include "globals.h"
 #include "time.h"
 
+// Escape a string for safe embedding in JSON string values.
```
ktomy, line 17: I don't think we need sanitization mechanism for now, but if you want it to exist, maybe we should move it to a library somewhere, or even better, import an existing library (made by some security professionals)

ktomy (changes requested):
I agree with UI changes, but I don't really see the sense for the JSON sanitization as we are not dealing with potentially offensive user input

ktomy:
Partially applied in https://github.com/ktomy/nightscout-clock/commit/7732b7a: the Dexcom credential UX changes in `data/index.html` have been applied to `main`. The `src/ServerManager.cpp` changes were not included.

Closing this PR because the author did not proceed with the requested changes. Thank you for the contribution.

PR #167, opened 2026-06-20
Fix mmol/L delta showing false ±0.1 for readings that display the same

martinsuchan:
## Problem

The value/diff clock face can show a delta of `-0.1` or `+0.1` mmol/L
between two readings that display as the *exact same number*.

Example: consecutive readings of 193 and 192 mg/dL both round to
10.7 mmol/L on screen, but the delta showed `-0.1` because it was
computed from the raw 1 mg/dL difference and converted to mmol/L
afterward:

193 mg/dL -> 10.722 -> displays "10.7"
192 mg/dL -> 10.667 -> displays "10.7"
raw diff = -1 mg/dL -> -0.056 mmol/L -> rounds to "-0.1"  (shown, but wrong-looking)

## Fix

`BGDisplayFaceValueAndDiff::getDiff()` still finds the min/max/last SGV
and runs the existing noise ("too many changes in last 6.5 minutes")
and magnitude ("diff too big") guards exactly as before, entirely on
raw mg/dL values — display unit never affects which readings are
considered valid or how large a jump is allowed.

Only the final rendering step changes: instead of converting the raw
mg/dL diff directly, it now diffs the two *displayed* values (each
independently rounded to the same precision shown on screen, via
`BGDisplayFaceTextBase::toDisplayTenths`/`formatDisplayTenths`, shared
with the existing single-value display code). Two readings that show
the same number on screen now always show a `0.0` delta.

`mg/dL` mode is untouched — the conversion is the identity function
there, so behavior for mg/dL users is byte-for-byte unchanged.

## Trade-off (intentional)

Because this diffs two independently-rounded values instead of
rounding the true underlying difference once, the delta can differ
from "true rate of change, rounded once" by up to 0.1 mmol/L for
larger gaps — e.g. a 101 -> 190 mg/dL change (true delta 4.94 mmol/L)
now shows `+5.0` instead of `+4.9`. This is bounded and intentional:
the delta is defined as "what changed between the two numbers you can
see," not the precise instantaneous rate of change.

## Testing

Not tested on hardware. Verified by manual trace and by simulating the
exact logic (min/max/base/threshold/render) against several cases,
including the reported bug (193/192 mg/dL), a larger gap showing the
documented trade-off, and confirming mg/dL-mode output is unchanged
from before this PR.

### DIFF
--- a/src/BGDisplayFaceValueAndDiff.cpp
+++ b/src/BGDisplayFaceValueAndDiff.cpp
@@ -63,7 +63,7 @@ String BGDisplayFaceValueAndDiff::getDiff(const std::list<GlucoseReading>& readi
         foundReadings.resize(2);
     }
 
-    // cucle through found readings, get min and max SGVs
+    // cycle through found readings, get min and max SGVs
     int minSGV = 9999;
     int maxSGV = 0;
     for (const auto& reading : foundReadings) {
@@ -74,7 +74,9 @@ String BGDisplayFaceValueAndDiff::getDiff(const std::list<GlucoseReading>& readi
             maxSGV = reading.sgv;
         }
     }
-    int base = last.sgv;
+    minSGV = getComparableSGV(minSGV);
+    maxSGV = getComparableSGV(maxSGV);
+    int base = getComparableSGV(last.sgv);
     if (minSGV != base && maxSGV != base) {
 #ifdef DEBUG_DISPLAY
         DEBUG_PRINTLN("Too many changes in last 6.5 minutes");
@@ -84,7 +86,7 @@ String BGDisplayFaceValueAndDiff::getDiff(const std::list<GlucoseReading>& readi
 
     int diff = minSGV == base ? base - maxSGV : base - minSGV;
 
-    if (std::abs(diff) > 99) {
+    if (std::abs(diff) > getComparableSGV(99)) {
 #ifdef DEBUG_DISPLAY
         DEBUG_PRINTLN("Diff is too big: " + String(diff));
 #endif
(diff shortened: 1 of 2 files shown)


src/BGDisplayFaceValueAndDiff.cpp
```diff
@@ -96,11 +98,29 @@ String BGDisplayFaceValueAndDiff::getDiff(const std::list<GlucoseReading>& readi
...
 
 #ifdef DEBUG_DISPLAY
     DEBUG_PRINTF("SGV Diff: %s (%d readings)", diffString.c_str(), foundReadings.size());
 #endif
 
     return diffString;
 }
+
+int BGDisplayFaceValueAndDiff::getComparableSGV(int sgv) const {
+    if (SettingsManager.settings.bg_units == BG_UNIT::MGDL) {
```
ktomy, line 111: Internally sgv is ALWAYS in mgdl, there is no chance it is in mmol. If for a reasoin it is in mmol, this has to be fixed on the data source level, not here

src/BGDisplayFaceValueAndDiff.cpp
```diff
@@ -84,7 +86,7 @@ String BGDisplayFaceValueAndDiff::getDiff(const std::list<GlucoseReading>& readi
 
     int diff = minSGV == base ? base - maxSGV : base - minSGV;
 
-    if (std::abs(diff) > 99) {
+    if (std::abs(diff) > getComparableSGV(99)) {
```
ktomy, line 89: This has no sense, what if setting is mmol/l? then the max difference is 99/1.8 mmol?

src/BGDisplayFaceValueAndDiff.cpp
```diff
@@ -96,11 +98,29 @@ String BGDisplayFaceValueAndDiff::getDiff(const std::list<GlucoseReading>& readi
...
+
+int BGDisplayFaceValueAndDiff::getComparableSGV(int sgv) const {
+    if (SettingsManager.settings.bg_units == BG_UNIT::MGDL) {
+        return sgv;
+    }
+
+    return round((float)sgv / 1.8);
+}
+
+String BGDisplayFaceValueAndDiff::getPrintableDiff(int diff) const {
```
ktomy, line 118: I can't see any difference of this method and `BGDisplayFaceTextBase::getPrintableReading`, am I wrong? Why did you introduce it?

src/BGDisplayFaceValueAndDiff.cpp
```diff
@@ -96,11 +98,29 @@ String BGDisplayFaceValueAndDiff::getDiff(const std::list<GlucoseReading>& readi
...
 
     return diffString;
 }
+
+int BGDisplayFaceValueAndDiff::getComparableSGV(int sgv) const {
+    if (SettingsManager.settings.bg_units == BG_UNIT::MGDL) {
+        return sgv;
+    }
+
+    return round((float)sgv / 1.8);
```
ktomy, line 115: I don't understand this 1.8 , I did not test, but this seems very wrong. If you thought of conversion from mgdl to mmol, this should be /18 , but then in `getPrintableDiff` it is divided again.
Look, this line is the one which avoids the "false +/- 0.1" but it does it with the price of showing the difference wrong for any values bigger then 0.2 , at least this is how I understand it

ktomy (changes requested):
(without actually testing this change on the device), I think this PR does not solve the problem, treats symptoms (shows 0 when there difference in BG in mg/dl is +/- 1 , but calculates difference in a wrong way for other values.

martinsuchan:
Hi, updated the PR with better description what is the problem, and proposed fix. Also simplified the code change to only touch the final rendering step. Instead of converting the raw mg/dL diff directly, it now diffs the two displayed values.

PR #169, opened 2026-07-25
Move machine-specific PlatformIO upload settings to git-ignored local override

0m4r:
See commit. Moves upload_port/upload_speed to an untracked platformio.local.ini via extra_configs; scripts read it first and fall back to platformio.ini; docs updated. Enables building/flashing on any machine without editing tracked files.


0m4r:
Opened against the wrong repo by mistake; reopening against the fork 0m4r/nightscout-clock.

PR #170, opened 2026-07-25
Add four clock faces (Diagnostics, Battery/Uptime, Rainbow, Smiley), display performance/stability fixes, and per-machine PlatformIO config

0m4r:
## Summary

This branch adds four new clock faces, a set of display performance and stability
fixes, and a per-machine PlatformIO configuration workflow so contributors don't
have to edit tracked files to flash their own board. It builds on `ulanzi_debug`
(`pio run -e ulanzi_debug`) and has been verified on hardware (Ulanzi TC001).

## PlatformIO configuration

- **Per-machine overrides via `extra_configs`** — `platformio.ini` now merges an
  untracked `platformio.local.ini` on top of the tracked config, so your serial
  port and upload speed never end up in a commit. (`platformio.ini`)
- **New template** documenting `upload_port` / `upload_speed` per OS; copy it to
  `platformio.local.ini` to get started. (`platformio.local.ini.example`)
- **Safer defaults** — `upload_speed` lowered to a conservative `115200` (override
  to `921600` locally once stable), and the hard-coded `upload_port` removed so
  PlatformIO auto-detects the port. (`platformio.ini`)
- **Ignore the local override** so it's never committed. (`.gitignore`)
- **Helper scripts** read the local override first and fall back to
  `platformio.ini`; added `--fs` and `--all` flows for filesystem/full-device
  flashing. (`scripts/build.sh`, `scripts/upload.sh`, `scripts/reset.sh`)
- **Emulator usage docs** — a docstring covering prerequisites, CLI modes, handy
  test values, and a curl equivalent for driving the device's API source.
  (`scripts/ns_emulator.py`)

## Code changes

**Display performance**
- Removed the redundant blank-frame blit on every refresh (each WS2812
  `matrix->show()` is a ~7–8 ms interrupts-disabled blit), eliminating a visible
  flash and wasted work. (`src/BGDisplayManager.cpp`, `src/DisplayManager.cpp`)
- Diagnostics/text faces refresh at a frame rate only while content actually
  scrolls, and fall back to a static, idle per-minute refresh when it fits.
  (`src/BGDisplayFaceDiagnostics.cpp/.h`, `src/BGDisplayFaceTextBase.cpp/.h`)
- Cache the date/time string off the frame hot path
  (`src/BGDisplayFaceDiagnostics.cpp/.h`); single map lookup in `getTextWidth`
  (`src/DisplayManager.cpp`); unit-specific refresh-time tracking
  (`src/BGDisplayManager.cpp/.h`).

**Display correctness / shared state**
- Reset the global font on every face switch so faces that render text without
  pinning a font no longer inherit a leaked LARGE font and clip; added an
  `onActivate()` hook to reset per-face view state on (re)selection, including the
  boot/default face. (`src/BGDisplayManager.cpp`, `src/BGDisplayFace.cpp/.h`,
  `src/BGDisplayFaceDiagnostics.cpp/.h`)
- Pin the compact font for standalone/status messages so they stay centered
  (`src/DisplayManager.cpp` fatal errors, `src/BGSource.cpp` "To API" prompt);
  fixed a missing `break` in `setFont`. (`src/DisplayManager.cpp`)

**Peripherals**
- Low-battery beep no longer cuts off an active melody; lowered buzzer volume.
  (`src/PeripheryManager.cpp`)

## Docs

- Documents the new Smiley face and the `ns_emulator.py` testing flow. (`README.md`)
- Notes the `platformio.local.ini` workflow and build/flash scripts.
  (`CONTRIBUTING.md`)
- Repository guidelines updated for the local-override config and emulator.
  (`AGENTS.md`)

## New screens

Four new clock faces, registered in `src/BGDisplayManager.cpp/.h` and selectable
via the default-clock-face dropdown in the web UI (`data/index.html`):

- **Diagnostics** (`6`) — scrolling date/time alongside the BG value and trend
  arrow; scrolls only when the content overflows, otherwise static.
  (`src/BGDisplayFaceDiagnostics.cpp/.h`)
- **Battery and uptime** (`7`) — device battery percentage and uptime.
  (`src/BGDisplayFaceBatteryUptime.cpp/.h`)
- **Rainbow big text** (`8`) — the large BG value cycling through rainbow colors.
  (`src/BGDisplayFaceBigTextRainbow.cpp/.h`)
- **Smiley** (`9`) — a round mood face beside a big centered reading: happy/green
  in range, sad/blue below the low limit, angry/red above the high limit, and a
  grey neutral face for no/stale data. Wide mmol/L readings fall back to a compact
  font so they never clip. (`src/BGDisplayFaceSmiley.cpp/.h`)


ktomy:
Please split this PR in a few more human-reviewable PRs. In this way I'll be able to review it. I would propose the following PRs split:
- Adding new faces (this will be the simplest to review and has the best chances to be imported as it does not change things and only adds new faces). Please don't include flickering-avoiding logic or anything solving other issues
- Adjustments to build process, everything related to platformio.ini, build scripts and environment. Please adjust the ns_emulator script as now part of the "smiley face" logic testing leaked in that script. Another adjustment I would propose is related to the IP address. As you are making the port configurable, do the same for the IP address, I can bet your clock is not on 192.168.86.24 :)
- Screen refresh adjustments: this one is the most difficult one, I tried to fix this stuff many times, but it always had side effects because of now different the data sources post values. So it has to be heavily tested, on all the production sources - dexcom, libre, nightscout (and maybe even medtrum). Every time I fixed it, what was happening is that it was working great for one or two sources, but was leaving black screen for under a minute for other sources

ktomy:
Sorry, now your PR has a lot of conflicts as I implemented yet another trial of making screen refresh better.

Also, I really liked smiley face and the idea of low battery alert, even if you will not touch your PR I will implement them anyway.

But if you decide to address my comments, please also do something with fonts handling, this font changes leaked everywhere, and even if my BigText changed the font, what you can do is to create an event for all the faces which is triggered when the face is changed, so that each face can clean after itself, and change font there... or something... I really did not understand why so much attention (code) was dedicated to the font change

0m4r:
Hey! Thanks for the feedback!
I am on holidays those days, I will be able to look into this again around end of August!

Let me know what's left! 😜

0m4r:
@ktomy I have resolved the conflicts for now.
As I am really a beginner in C++, maybe you could check if I have done it right (I have got some help from AI, but I tried to understand it!).

I will need some more time to split the changes into individual parts.

Thanks and keep up the good work!

PR #177, opened 2026-09-05
Make the data-is-old colour configurable; gray is invisible at minimum brightness

nishanm:
Raised from the investigation in #176.

## The bug

`COLOR_GRAY` is `0xa514` â€” `(165, 162, 165)`. **No channel is at its maximum**, unlike every other colour the firmware uses. At `MIN_BRIGHTNESS` the panel cannot render it.

Measured on a Ulanzi TC001 running `brightness_mode: manual`, `brightness_level: 1`:

| Reading age | On the panel |
|---|---|
| fresh | `120`, green |
| 7 min | `120`, cyan |
| **25 min â€” past the data-is-old threshold** | **nothing at all** |

The cyan row was ten seconds earlier at the same brightness on the same reading, so this is not the display being off or the value failing to render â€” it is gray specifically.

**The consequence: on a clock running dim, a fully stale reading shows as a blank display rather than a gray number.** That is indistinguishable from the clock being switched off or unplugged. The state that most needs to be noticed is the one that cannot be seen.

It is not limited to `manual` level 1. `AUTO_DIMMED` pins the panel to `MIN_BRIGHTNESS` whenever the room is below 1 lux, so anyone using auto brightness in a dark room hits this every night â€” and has no reason to suspect a colour constant.

## The change

A new `data_old_color` setting offering `gray`, `cyan`, `magenta` and `blue`. The three alternatives all keep at least one channel at maximum and remain visible at brightness 1.

**An existing clock is unchanged.** A config without the key falls back to `GRAY`, so nothing moves under anyone who does not touch the setting.

**A new clock is not.** `config_initial.json` seeds `cyan`, and that file is only read on a filesystem with no config yet. Leaving it at `gray` would have meant a brand new device still shipping with the bug this PR exists to fix, escapable only by finding a setting you have no reason to look for. The split is deliberate: the seed decides what a fresh install gets, the C++ fallback decides what an upgrade keeps.

The colour is resolved once in `BGDisplayFace::getDataOldColor()`, so every face that shows the old state picks it up: the value and trend arrow on the text faces, the delta on Value and diff, and the age bars — and the "no data" screen on both `BGDisplayFace` and `BGDisplayFaceClock`, which was drawn in the same unreadable gray.

`enums.h` gains a single `DISPLAY_COLOR` listing every colour the firmware draws with, and the choice is restricted in the WebUI rather than in firmware — so adding a colour to the dropdown needs no firmware change. The enum value **is** the RGB565 code, so there is no mapping table and `getDataOldColor()` is a `static_cast`; the `COLOR_*` macros in `globals.h` are now defined from it, which puts the codes in exactly one place instead of two. Every existing call site is unchanged.

`BG_COLOR_OLD` is removed: it would have been a way to get the old colour while bypassing the setting.

### Since your review

- **The rename is backward compatible.** `stale_old_color` was this setting's key before the review renamed it, so a config written by an earlier build carries it and no `data_old_color`. The loader reads the new key, falls back to the old one, and only then to `GRAY` — without it a user who had already configured `blue` to escape this bug would have been put silently back on gray by the update. The settings page reads the same chain, and drops the old key when it saves, so the migration completes rather than leaving both.
- **The colour is restricted in firmware as well as the WebUI.** `DISPLAY_COLOR` lists every colour the firmware draws with, which is wider than what this setting should accept: `BLACK` is invisible, red, green and yellow are what the glucose bands mean, and white is the fresh trend arrow. `isDataAgeColor` holds the setting to the four the dropdown offers and says so in the log when it refuses one, so a hand written config cannot reintroduce the failure.
- **`showNoData()` now sets its font.** It never did, so it inherited whatever the last face used — on Big text that is `FONT_TYPE::LARGE`, which renders "No data" far wider than 32 pixels and clipped it to "O DAT". That predates this PR by a long way and nobody had reported it because the screen drew in gray and was invisible; fixing the colour without the font would have made a broken screen visible and looked like this PR caused it.
- **The setting sits in Device settings, under Brightness level.** It was in Advanced settings, which is collapsed by default, two cards away from the brightness control that sends people looking for it. Its help text already refers to running the clock dim.
- Spelling normalised to `color` throughout, matching the rest of the project.

## Verified on hardware

Same panel, same brightness (`manual` level 1), same 25-minute-old reading:

- **before:** nothing displayed
- **after (`data_old_color: blue`):** `120` displayed in blue

The reworked version has since been re-flashed and re-checked, against a clock whose config genuinely predated the rename — `stale_old_color: "blue"`, no `data_old_color`:

| State | On the panel |
|---|---|
| 25 min old, fallback resolving the old key | `120` in **blue**, trend arrow blue too |
| same, with `data_old_color` hand-set to `"black"` | still **blue** — the subset check refused it |
| same, setting swapped to `magenta` | **magenta**, so the colour follows the setting rather than happening to look right |
| no readings at all | "No data" in the same colour, legible, in both blue and magenta |

A factory-fresh filesystem was then uploaded to confirm the seed: `data_old_color` comes up as `cyan` on a clock with no prior config.

Measured colour brightness order at `MIN_BRIGHTNESS`, observed on the panel: **cyan > gray > blue**. Blue is the dimmest of the three alternatives but clearly legible, which suits a terminal "ignore this number" state.

## Deliberately not changed

`DisplayManager::showFatalError()` also draws in gray and is likely invisible at brightness 1 for the same reason. It is not a display face so it cannot reach `getDataOldColor()`, and "something broke" is a different state from "the data is old" — worth its own change rather than being folded in here.

(An earlier revision of this PR also left `showNoData()` alone. That was the wrong call and it is now fixed: "no data" is the same class of problem — drawn in the same gray, so at low brightness the screen telling you the clock has no readings was itself invisible.)

---

ðŸ¤– Generated with [Claude Code](https://claude.com/claude-code)


ktomy:
I think there is no sense in STALE_COLOR enum, just introduce an enum with all the colors used in the app and restrict the selection from the WebUI. This would simplify the code.
Actually there is one more thing to change, there is the "NO DATA" situation, I think your PR does not handle this. The color of "NO DATA" has to be the same as the old data. also I would rename `getOldColor` to `getDataOldColor`

nishanm:
Thanks — all three done, pushed as a follow-up commit so you can see exactly what changed since your review.

**1. `STALE_COLOR` is gone.** `enums.h` now has a single `DISPLAY_COLOR` listing every colour the firmware draws with, and the dropdown is restricted in the WebUI instead of in firmware — so adding a colour to the choice needs no firmware change at all.

I took it one step further than a straight rename, and it's the part most worth a second opinion: the enum value **is** the RGB565 code, so there's no mapping table — `getDataOldColor()` is now just a `static_cast`, and the whole `colorFor()` switch is deleted. The `COLOR_*` macros in `globals.h` are defined from the enum, which puts the codes in exactly one place instead of two lists that can drift. Every existing call site is untouched.

I also dropped `BG_COLOR_OLD`, since it would have been a way to get the old colour while bypassing the setting.

**2. "NO DATA" now follows the same setting**, on both `BGDisplayFace::showNoData()` and `BGDisplayFaceClock::showNoData()`. You're right that it belonged in this PR — it's the same class of problem, drawn in the same unreadable gray, which means at low brightness the screen telling you the clock has no readings was itself invisible. Leaving it out was the wrong call on my part.

**3. `getOldColor()` → `getDataOldColor()`.**

One judgement call I'd like your view on: I let the setting and its JSON key follow that rename too, `stale_old_color` → `data_old_color`, so it matches `getDataOldColor()` and the existing `bg_data_too_old_threshold_minutes`. Nothing has shipped with the old key, so this seemed like the moment — but it's a one-line revert if you'd rather keep it as it was.

**Still not changed:** `DisplayManager::showFatalError()` also draws in gray and is probably invisible at brightness 1 for the same reason. It isn't a display face so it can't reach `getDataOldColor()`, and "something broke" is a genuinely different state from "the data is old", so I'd rather it were its own change than folded in here. Happy to do it separately if you want it.

**Status:** builds clean for `ulanzi`. The PR is now 13 files, +131/−15, and the firmware side is smaller than before your review despite doing more.

One thing I should be straight about: **this rework has not been back on hardware.** The original version was verified on a TC001 (the 25-minute-old reading that displayed nothing now displays in blue at the same brightness), but that was the `STALE_COLOR` implementation. The refactor is behaviour-preserving by construction and compiles, and the "no data" change is new code that has only been reasoned about. Say the word and I'll flash it and report back before you merge — I'd rather you had that than take my word for it.

**#178 is rebased on top of this** and updated to the same API, so it still applies cleanly and stays a one-commit diff once this lands. It keeps its own `stale_early_*` naming for now — if you'd prefer those aligned with `data_old_color` too, that's easy, just say which way you want it to read.

nishanm:
## Hardware pass done — on a real TC001, at `brightness_level: 1`

Following up on my note that the rework hadn't been on a device. It has now.

### The "NO DATA" change is verified, with a control

Just seeing the right colour once isn't worth much, so I tested it as a swap rather than an observation:

| `data_old_color` | What the panel showed in the no-data state |
|---|---|
| `blue` | no-data text in **blue** |
| `magenta` | same state, same text, now **magenta** |

The text follows the setting, so `showNoData()` is demonstrably reading `getDataOldColor()` rather than happening to look right. Previously this was hardcoded `COLOR_GRAY`, which at brightness 1 renders as nothing at all.

### ⚠️ Reaching `showNoData()` is harder than it looks — worth knowing if you test this

My first two attempts didn't test the change at all, and both looked plausible:

1. **The post-boot window is not it.** `BGSource::tick()` clears the matrix and prints `"To API"` while `firstConnectionSuccess` is false, and that path runs *instead of* the render. So the gap between boot and the first reading shows `"To API"`, never `"No data"`.
2. **Switching to the API source with no entries is not it either** — the first connection never succeeds, so it parks on `"To API"` indefinitely.

What actually works: API source → POST one entry (`updateReadings()` sets `firstConnectionSuccess`, and the number renders) → `DELETE /api/v1/entries` (clears readings and sets `hasNewDataFlag`, forcing a render) → that render has zero readings and calls `showNoData()`.

### Also verified

- **The refactor runs.** Boots, connects, serves the UI, no regressions seen.
- **The setting round-trips.** After a save, `config.json` contains `data_old_color` and `stale_old_color` is gone — which only the new build does, so this is functional proof the reworked firmware is what was running, not just that it booted.
- **Firmware-only upload preserves config** (all 66 keys, WiFi and source credentials intact).

### Not verified this pass

The value and trend arrow in the data-is-old colour on a genuinely stale reading — that needs a real 20-minute gap and I didn't manufacture one. It was verified on hardware in the original version of this PR, and the refactor routes both through the same `getDataOldColor()`, but I'd rather say so than imply more coverage than I have.

### One thing the fix exposed: `"No data"` is clipped

Now that the text is actually visible, it's clear it doesn't fit. On the 32x8 panel `printText(0, 6, "No data", TEXT_ALIGNMENT::CENTER, 0)` renders as roughly `o dat` — the leading `N` and trailing `a` fall off both ends.

This is pre-existing and nothing to do with this PR; it was simply invisible before, so there was no way to notice. Happy to open it separately — a shorter string would fix it, though I'd rather you picked the wording than have me guess.

### And a data point on the key rename

I flagged earlier that renaming `stale_old_color` → `data_old_color` was a judgement call. Flashing this over a device that already had the old key set showed the cost concretely: the config still said `stale_old_color: "blue"`, the new firmware looked for `data_old_color`, found nothing, and **fell back to gray** — silently reintroducing exactly the invisible-stale bug this PR fixes, on the one device that had it configured.

Nothing has shipped with either key so only my own clock was affected, but if you'd prefer, the safe version is a one-line fallback: read `data_old_color`, else `stale_old_color`, else gray. Or just tell me to revert the rename. Your call — I have no attachment to it.

nishanm:
## Testing complete — including the gap I flagged, and a negative control for the bug itself

This closes the "not verified this pass" caveat from my previous comment. All of it on one TC001 at `brightness_level: 1`, one state after another, minutes apart.

Readings were driven by switching to the API source and posting a single backdated entry, which makes every age band reachable on demand. Thresholds on the device: early 6 min, data-is-old 20 min (default).

| Reading | `data_old_color` | On the panel |
|---|---|---|
| fresh, 1 min | — | **green** (in-range band) — control |
| same value, 8 min | — | **cyan** (early stale) |
| same value, 25 min | `blue` | **blue, clearly readable** |
| **same value, 25 min** | **`gray`** | **nothing at all — blank panel** |
| no readings | `blue` | no-data text **blue** |
| no readings | `magenta` | no-data text **magenta** |

### The fourth row is the one that matters

That is the bug, reproduced under control today rather than quoted from an earlier session: **the same 112, the same age, the same brightness, seconds apart — readable in blue, invisible in gray.** Nothing else differed. At `MIN_BRIGHTNESS` a stale reading in the shipped default colour is indistinguishable from the clock being switched off.

### The value and arrow are covered now, not just "no data"

My earlier comment verified `showNoData()` but not the reading itself. Row 3 covers it: the value and trend arrow both render in the configured colour via `getDataOldColor()`.

And the colours are shown to *follow the setting* rather than merely looking right — rows 5 and 6 are the same state with only the setting changed, and the text changed with it.

### Ordering holds

Rows 1→2→3 are the same reading at three ages, and the precedence behaves: glucose band, then early-stale, then fully-old. No state got stuck or skipped.

### Unchanged from before

- `"No data"` is clipped to roughly `o dat` on the 32x8 panel. Pre-existing, purely cosmetic, and only noticeable now that the text is visible at all — flagging it rather than proposing a fix, since it's the panel width and the wording is yours to choose.
- The `stale_old_color` → `data_old_color` rename still needs your call: leave it, add a one-line fallback, or revert. I have no attachment to it.

nishanm:
## Added the fallback rather than leaving the rename to chance

Following up on my own flag about `stale_old_color` → `data_old_color`. Rather than ask you to pick, I've taken the option that cannot produce a bad state:

```cpp
settings.data_old_color = displayColorFromString(
    (*doc)["data_old_color"].as<String>(),
    displayColorFromString((*doc)["stale_old_color"].as<String>(), DISPLAY_COLOR::GRAY));
```

Read the new key, fall back to the old one, and only then to `GRAY`.

### Why it's worth the extra call

Without it, a config written by an earlier build carries only the old key, so the firmware defaulted straight to gray on update — **silently putting that user back on the one colour the panel cannot render at `MIN_BRIGHTNESS`.** And the people it would hit are precisely those who had already run into the bug and configured their way out of it. That's the worst possible group to regress.

### Verified on hardware, both directions

Same TC001, `brightness_level: 1`, same reading, same age — only the firmware differing:

| Config on device | Build | 25-minute-old reading |
|---|---|---|
| `stale_old_color: "blue"`, no `data_old_color` | before this commit | **blank panel** (fell back to gray) |
| `stale_old_color: "blue"`, no `data_old_color` | with this commit | **blue, readable** |

The device was booted with the new key genuinely absent — confirmed by re-reading `/config.json` after the save — so the fallback is what resolved the colour, not a leftover value.

### It's self-healing

The write path is untouched, so the first save normalises the config to `data_old_color` and the fallback is never consulted again. It's a migration shim, not a permanent second code path.

#178 is rebased on top and still applies cleanly.

nishanm:
Pushed a few more changes and updated the description. Summary of what moved, since none of it is in the part you already read:

**The rename needed a migration.** `stale_old_color` was the key before your review renamed it, so a config written by an earlier build has it and no `data_old_color`. The loader now reads the new key, falls back to the old one, and only then to `GRAY`. Without that, anyone who had already configured `blue` to escape this bug would have been put silently back on gray by the update — the people most affected being exactly the ones who had already worked around it. The settings page reads the same chain and drops the old key on save.

**The choice is restricted in firmware too.** `DISPLAY_COLOR` covers every colour the firmware draws with, which is wider than this setting should take — `BLACK` is a blank panel, red/green/yellow are the glucose bands, white is the fresh trend arrow. `isDataAgeColor` holds it to the four the dropdown offers and logs what it refused. The old `STALE_COLOR` carried that constraint in the type; folding it into one enum was right, but it widened the domain, so the check moved rather than disappeared.

**`showNoData()` never set a font.** It inherited whatever the last face used, and on Big text that is `FONT_TYPE::LARGE` — "No data" rendered wider than 32 pixels and clipped to "O DAT". That is not new and this PR did not cause it; it has been invisible because the screen drew in gray. But making the colour configurable without the font would have made a broken screen visible and looked like this PR broke it, so it is fixed here.

**One behaviour change worth your opinion:** `config_initial.json` now seeds `cyan` instead of `gray`. An upgrade is still unaffected — a config with no key hits the `GRAY` fallback and nothing moves — but a brand new clock was otherwise still shipping with the bug, escapable only by finding a setting there is no reason to look for. If you would rather a fresh install stayed gray, that is a one line revert of the seed and nothing else changes.

Also moved the setting from Advanced settings into Device settings under Brightness level, since it exists because of brightness and Advanced is collapsed by default, and normalised the spelling to `color` to match the rest of the project.

**Re-tested on a TC001 at `brightness_level: 1`**, on a clock whose config genuinely predated the rename. Same reading and brightness throughout:

| State | Panel |
|---|---|
| 25 min old, fallback resolving the old key | `120` blue, arrow blue |
| `data_old_color` hand-set to `"black"` | still blue — refused |
| swapped to `magenta` | magenta |
| no readings | "No data", legible, both colours |

Then a filesystem upload to confirm the seed on a factory-fresh device: it comes up `cyan`.

PR #178, opened 2026-09-05
Add an early-stale color option to the Big text face

nishanm:
Big text has no age bars, so a missed reading doesn't show until the old-data colour takes over at 20 minutes. As discussed below, Big text itself can now colour the value after a missed reading, from a panel of its own:
- Late colour Off (default), cyan, blue or magenta, after 6, 10 or 15 minutes. The old-data colour still takes over at the data-is-old threshold.
- The page refuses a late colour equal to the old-data colour, and minutes at or past the no-data time.
- This adds the per-face panel ("drawer") to the active faces block from #195: collapsed by default, shown only for an active face that has settings. Big text is its first user.

The global `stale_early_*` settings from the earlier version are gone. Depends on #195 (which depends on #192); only the last commit is new.


nishanm:
## Hardware pass — the early-stale indicator works on a real TC001

This PR is stacked on #177 and has been rebased onto the reworked version of it, so this build exercises both. Tested on a TC001 at `brightness_level: 1`, face 3 (Big text), with `stale_early_minutes: 6` and the default 20-minute data-is-old threshold.

The same reading was replayed at three ages by posting a single backdated entry to the API source:

| Reading age | On the panel |
|---|---|
| 1 min | **green** — in-range glucose band (control) |
| 8 min | **cyan** — early stale |
| 25 min | **blue** — fully old |

So the three states are distinguishable on a face that has no age bars, and the precedence in `showReading()` behaves as intended: fully-old wins over early-stale, which wins over the glucose band. Nothing stuck or skipped a band.

The control row matters as much as the others — a fresh reading stays green, so the indicator isn't simply colouring everything.

### Why this is worth having

On Big text there is no room for age bars, so today a reading looks completely normal right up until it crosses the 20-minute threshold. Green means "in range"; it does not mean "current". The gap between *one* missed reading and *four* missed readings is the part that changes what you do, and until it fires there is no signal at all.

That's the whole feature: at 8 minutes the number stops claiming to be fresh.

### Status

Still defaults to **disabled**, so it is a no-op unless someone turns it on. Rebased cleanly on the reworked #177 and stays a one-commit diff once that lands. Builds clean for `ulanzi`.

nishanm:
Rebased onto the reworked #177 and pushed some corrections. Description updated; the substance:

**A missing key loaded 0, not the declared 6.** `Settings.h` says `stale_early_minutes = 6`, but `as<int>()` on an absent key gives 0, so a config written before this feature loaded a threshold that the header, the page and the file all disagreed about. Invisible in practice only because the feature also defaults to off.

The fix is worth a second look because the tidier version is wrong. `operator|` returns the default whenever `is<T>()` is false, and the page writes this field with `.val()`, so a real `config.json` holds the **string** `"6"` — `| 6` would have ignored every value a user had ever set and quietly loaded 6. It tests `isNull()` instead, keeping `as<int>()`, which parses the string. I only caught it by reading a config off a running clock rather than reasoning about it, and then confirmed on the device: with `"12"` stored as a string, 8 minutes stayed green and 15 minutes went cyan. With `| 6` both would have been cyan.

**`isEarlyStale` now checks both ends itself.** It only tested the lower bound and relied on callers checking fully-old first. Both call sites did, and the comment said so, but that puts the correctness of a colour in the call order of whoever writes the next face — and the failure mode is the milder state winning over the worse one.

**`stale_early_color` goes through the same restricted parser** as the data-is-old colour, so neither can be set to `BLACK` or to a glucose band colour. Worth flagging that this hunk merged cleanly during the rebase while the API underneath it had moved, so it needed checking rather than trusting the conflict list.

Smaller: the page writes the threshold with `parseInt` so the file holds a number like `face_cycle_interval_seconds` does; the row now has a rule above it like every other row on the card; its description moved inside the block that hides when the switch is off, since it was the only row showing details while disabled; spelling normalised to `color`.

Re-verified on hardware after the rebase — same value, same brightness, age the only variable: 1 min green, 8 min cyan, 25 min blue.

nishanm:
Found a real bug in this PR after running it on a clock for a few hours, and it is one you would have hit immediately.

**The early stale color was unreachable in normal use.** `getRenderDecision` returns `NONE` on a time tick unless `dataIsOld` has changed. Early stale is a time-driven transition too — a reading crosses it while sitting on screen, with no new data to trigger a redraw — so a value drawn green when it arrived stayed green as it aged, and only repainted when it crossed the *later* fully-old threshold, jumping straight to the data-is-old color.

It only affected faces inheriting `BGDisplayFace` directly: Big text, Simple, Value and diff. Faces inheriting `BGDisplayFaceWithAge` redraw every tick to advance their age bars, so they always showed it correctly. That is exactly the set of faces this PR exists for, since the premise is that Big text has no room for age bars.

Fixed by carrying the early stale state in `RenderContext` alongside `dataIsOld`, having the manager remember what it last rendered, and repainting on a tick when either changes. The rule itself moved into a free `isReadingEarlyStale()` so the render loop and the faces ask the same question instead of each keeping a copy.

**Why my earlier hardware results did not catch it, since that matters for how much the rest of my testing is worth:** every age test posted a fresh reading at each age, and a data arrival always forces a full redraw, so the color was recomputed from scratch every time. Those runs proved the color was *selected* correctly and never that it was *refreshed*. The failing axis was time passing with no new data — the normal case on a clock, and the one case I never exercised. The "1 min green / 8 min cyan / 25 min blue" table in the description was true and measuring the wrong thing.

Re-verified with a single reading posted and then left alone, thresholds compressed to 2 and 6 minutes to keep the run short:

| age | panel |
|---|---|
| 0.5 min | green |
| 2.5 min | **cyan**, nothing posted since — the transition that could not happen before |
| 8.7 min | blue |

Caught because someone running three of these noticed a number still green after a missed reading.

nishanm:
Based on your feedback on #177, the same changes are incorporated here: rebased onto main so the PR is back to just the early-stale commits (it had gone CONFLICTING after the merge), the `stale_old_color` shim and the firmware colour allow-list are gone with it, the early-stale colour is read with `displayColorFromString()` like `data_old_color` is now, the `isEarlyStale()` forwarder is removed in favour of calling `isReadingEarlyStale()` directly, and the comments are trimmed. Builds for ulanzi; no functional change.

nishanm:
One more small fix: data-is-old begins strictly after its threshold while early stale ended strictly before it, so at exactly the threshold second the reading fell back to its glucose colour. Early stale now runs up to and including the threshold.

ktomy:
Hi, I don't really agree with this PR. Here is why:
- On any face except full-graph and BigText it is clear when the data is older than 5 minutes, so it means a missed reading
- We ignore full-graph as I do not think someone would keep it as a permanent face (while I think it can be included in face rotations), so the only affected face is BigText.
- Instead of adding more settings and trying to generalize, why wouldn't you just add one more face, something like BigTextWithStaleIndicator and customize it as per your liking? I am ok if it would change its color on 5 or 6 minutes of missing data, even with non-standard colors, as soon as you document it in readme.

WDYT?

nishanm:
Fair on the scope: the age bars already say "missed reading" on every other face, so Big text is the only one hiding it. The setting was there to generalise, and I agree that isn't needed.

Where I'd push back is a new face. Faces don't compose: Big text, Big text (stale), Big text (dark), Big text (dark, stale) is where feature-as-face ends up, and the list every user scrolls past is already 6 and going to 8. A setting that defaults off costs the simple user nothing.

Middle ground with no setting and no face: Big text itself changes the value colour after one missed reading (fixed 6 min, fixed cyan), the same signal the bars give elsewhere in the only channel this face has, and the data-old colour still takes over at 20 min. Documented in the README. If you'd rather leave stock Big text untouched I'll do the separate face instead. Which do you prefer?

ktomy:
> Big text itself changes the value colour after one missed reading

Agree, but with configurable effect and timing. As we agreed to have settings per face, we can afford this

nishanm:
Sounds good. Once the Active faces block from #191 is in, I'll rework this into a Big text drawer: an early-stale colour (off, cyan, blue or magenta) and the minutes (6, 10 or 15), off by default so stock Big text is unchanged, and the data-old colour still takes over at its own threshold. The global settings go away.

nishanm:
Rebuilt as agreed: the setting lives in a Big text panel of the active faces block (#195), colour and minutes, off by default, and the global settings are gone. Description updated.

ktomy:
Rebased onto current `main` (`d4ba07c`) and resolved the conflicts.

- Dropped the older settings-page and active-face predecessor commits already integrated into `main`.
- Combined the Big text early-stale controls and validation with the current daily schedule and active-face UI, and preserved the stale-reading X documentation.
- Kept compressed web assets generated from source and shortened the new help text to stay within the existing 40 KB page budget.

Firmware and LittleFS builds pass, as do JavaScript syntax checks. Hardware behavior has not been tested.

nishanm:
One change on top of your rebase: the per-face drawers now sit with the face list, above the cycling switch, so a face's own settings are read where the face is chosen. Nothing about the early-stale option itself changed.

Firmware and LittleFS builds pass, and the page is 39,980 B against the 40,000 B budget.

PR #183, opened 2026-09-07
Add per-alarm alert windows with days and times

nishanm:
Proposed in #182, opened alongside it so the shape is easy to look at rather than to jump ahead of the discussion. Happy to change any of it, or to close this if you would rather the idea landed differently.

### What changed

Each alert (high, low, urgent low) can now be given any number of **alert windows**, each with its own days and its own start and end time. The alert sounds only while one of its windows is open.

**An alert with no windows alerts at any time**, so nothing changes for an existing user until they add one.

This replaces `alarm_<level>_silence_interval`, which was a single window applied to all seven days, expressed as a "do not alert" rule, and limited to `22_8` or `8_22`. It could not express "weekdays 09:00 to 17:00" or "every day 18:00 through to 08:00", which is what a clock with a limited useful window needs.

### Why it is safe to replace the old setting

Both old values are exactly the complement of an all-week window, so the migration is lossless and an upgraded clock alerts at precisely the same times as before:

| Old value | Meaning | Migrated to |
|---|---|---|
| `0` or empty | never silent | no windows, so alerts at any time |
| `22_8` | silent 22:00 - 08:00 | every day, 08:00 - 22:00 |
| `8_22` | silent 08:00 - 22:00 | every day, 22:00 - 08:00 |

Migration runs on load only when the window list is absent, and the superseded key is removed once windows are written, so it cannot be applied twice or come back later.

### Behaviour worth reviewing

- **Windows may wrap past midnight.** When the end time precedes the start time the window runs into the next day, and the days selected are the days it *starts* on. `BGAlarmManager` therefore checks yesterday's day mask as well as today's.
- **Days must be readable, not merely contain a readable digit.** `parseAlertWindowDays` now reports whether the whole value parsed and fills the mask only when it did, so an empty list, a stray character, a digit above 6 or the same day twice is refused by `/api/save` with a reason. It previously skipped anything it did not recognise and kept whatever bits were left, so `"Mon1"` saved successfully as Monday and the stored schedule was not the one submitted. `parseTimeOfDayMinutes` beside it has always been strict, and two parsers in one file disagreeing about how much they will guess seemed worth closing.
- **Anything unparseable is dropped rather than honoured.** A window with no days, an unreadable time, or zero length is discarded on load. A dropped window can only make the clock alert more, never less, which seemed like the only defensible direction here. The web UI will not save such a window, and `/api/save` now returns an error rather than accepting a schedule the clock will not keep.
- Windows are written even for a disabled alarm, so switching an alert off and saving does not throw away the schedule.

### Storage

```json
"alarm_low_alert_windows": [
  { "days": "12345", "from": "09:00", "to": "17:00" },
  { "days": "0123456", "from": "18:00", "to": "08:00" }
]
```

Days are `tm_wday` digits, so `0` is Sunday.

Writing the window list removes the superseded `alarm_<level>_silence_interval`, which makes the move one way: firmware old enough to predate windows would find neither key and alert at any time. That is the loud direction, but a schedule set here does not survive a downgrade, and the code now says so.

### One incidental change

`src/Settings.h` had no include guard. `BGAlarmManager.h` now needs the `AlertWindow` type, and including `Settings.h` from two places in one translation unit was an immediate `redefinition of 'class Settings'`. The guard is added in this PR because the change cannot build without it â€” please say if you would rather it were split out.

### UI

Each alert gets a list of rows, each a Mon-to-Sun toggle group plus a start and end time and a remove button, with an "Add window" button below and a validation message that names the specific problem. It reuses the existing Bootstrap components and adds no frontend dependencies. Happy to add screenshots if useful.

### How it was tested

**Now run on hardware** - full results in a comment below. On a real TC001 on Dexcom the migration ran against a genuine pre-existing config, and three schedules were checked with `-DDEBUG_ALARMS` so the decision is readable in the log rather than guessed at from whether a beep was audible. A closed window stayed quiet; a window wrapping midnight with only the wrong day ticked correctly stayed quiet; the same window with the right day ticked sounded. Every quiet result carries `glucose reading in alarm range` on the same ticks, so the silence is the window working rather than the alarm path being inert.

**The settings page has since been exercised too.** Driving the real page against the real firmware over the LAN: the Add window button is disabled while an alert is off and enables with it, rows add and remove, the day toggles collect in display order and are sorted before saving, and a Mon-Fri 02:00 - 03:00 window saved through the page came back after the restart as exactly `[{"days":"12345","from":"02:00","to":"03:00"}]`. The same save migrated all three legacy silence keys away. The filesystem has since been uploaded to that clock as well, so the page has now run from the device itself.

The malformed cases were posted at `/api/save` on the device: `"Mon1"`, `"17"`, `"11"`, an empty list and an eight character list were each refused with the reason, and re-reading `config.json` after all of them showed it untouched, so a rejected save writes nothing.

The firmware builds clean on `ulanzi_debug` (73.2% flash).

The web UI half was driven against a local stand-in for the device's web server, which serves the repo, hands out a crafted `config.json`, and captures what Save posts. That config mixes formats deliberately: the high alert arrives as an old `22_8` with no window list, the low alert as two windows including one that wraps midnight, and the urgent low as an empty list. Verified in the browser:

- the old `22_8` migrated to exactly one all-day 08:00 - 22:00 window
- both low windows rendered, including the wrapping one
- an empty list rendered no rows and means "alert at any time"
- clearing every day on a row raised the validation error and the save did not post, with a successful save afterwards as the control
- turning an alert off and saving kept its windows
- the superseded `alarm_<level>_silence_interval` key is gone from the posted config


nishanm:
Force-pushed a revised version of the same single commit (no review had happened yet, so I amended rather than stacking a fixup). Three changes, all from a self-review pass:

**An unknown time must not silence an alarm.** `getTimezonedTime()` returns its `struct tm` whether or not `getLocalTime()` succeeded — it logs and hands back the struct regardless. That is fine for drawing a face, but this feature was using `tm_wday` and `tm_hour` to decide whether an alarm may sound, so a clock that had not reached an NTP server could have had a schedule suppress a low reading based on an unset struct. There is now a `tryGetTimezonedTime(tm&)` that reports the failure, and `isInsideAlertWindow` treats an unknown time as "sound the alarm". This was the one finding I would call a real bug, and it contradicted the stated principle that nothing unreadable should ever be the reason an alert stays quiet.

**Stored times no longer reach the row markup.** The times were interpolated into `value="..."` when building each row. Everything else in that markup is owned by the page — day numbers and labels come from a constant, the row id from a counter — but the times come from `config.json`, so a malformed or hostile config could have broken out of the attribute. They are applied with `.val()` after the row is built instead. Verified with a config containing `">` as a stored time: it renders as a rejected field value, no element is created and nothing executes.

**Rendering a list validates once rather than once per row.** `renderAlertWindows` was calling the disable-sync and the validator after every appended row, each of which walks all rows. Row building is now separated from attaching, so a list is built and then checked a single time. Also renamed a jQuery `map` callback parameter that was called `day` while actually holding the checkbox element, and added comments for the two non-obvious decisions (why the day digits are sorted, and why `collectAlertWindows` still skips unusable rows when validation has already blocked the save).

Re-verified after the changes: the migration of an old `22_8` config, a window wrapping midnight, an empty list meaning "always", an invalid row blocking the save (confirmed by the save never posting, with a successful save afterwards as the control), a disabled alarm keeping its windows, and the superseded key being dropped from the saved config. Still builds clean on `ulanzi_debug`, and still not yet run on real hardware — that pass is still owed and I will report it on #182.

nishanm:
### Hardware pass done

Ran this on a real Ulanzi TC001 (0.30.0 + this branch), on Dexcom, reading 121 mg/dL at the time.

I uploaded **firmware only, not the LittleFS image**, on purpose: that leaves the device's existing `config.json` in place, so the migration ran against a real configuration written by the old build rather than one I had prepared. It also meant the clock never left the network.

**Migration, on a real pre-existing config.** All three alerts logged it at boot:

```
[loadAlertWindows]: No alert windows for alarm_urgent_low_alert_windows, migrating the silence interval instead
[loadAlertWindows]: No alert windows for alarm_low_alert_windows, migrating the silence interval instead
[loadAlertWindows]: No alert windows for alarm_high_alert_windows, migrating the silence interval instead
```

That device had `"0"` on all three, so it migrated to no windows, i.e. alerts at any time — unchanged behaviour, which is the point.

**Window evaluation.** I enabled the high alert at 100 so the live 121 reading sat inside the band, built with `-DDEBUG_ALARMS` so the decision is readable rather than inferred from whether I heard a beep, and drove the schedule through `/api/save` + `/api/reset`. Counts are ticks over roughly 45 seconds each:

| Window | Local time | Expected | Logged |
|---|---|---|---|
| every day `10:00 - 11:00` | Mon 03:2x | quiet | 56 × in range, 56 × `outside every alert window, staying quiet` |
| **Mon only** `18:00 - 08:00` | Mon 03:2x | quiet | 43 × in range, 43 × `outside every alert window, staying quiet` |
| every day `18:00 - 08:00` | Mon 03:2x | **sound** | 37 × in range, `Playing alarm sound (nee alarm occurred)`, then the repeat backoff |

The middle row is the one I most wanted on hardware. `18:00 - 08:00` wraps midnight, and at 03:2x on Monday the open half belongs to the window that *started* Sunday evening. With only Monday ticked it must stay shut, and it did. An implementation that checked today's mask against the whole wrapped range would have sounded there. The third row is the same window with every day ticked, so Sunday's mask is set and it correctly sounds.

Every quiet result carries `Alarms: glucose reading in alarm range` on the same ticks, so the silence is the window doing its job and not the alarm path being inert. Without that line the two quiet rows would prove nothing.

The device was returned to its original configuration afterwards and is back on Dexcom normally.

**Still not exercised on the device itself: the settings page.** I did not upload the filesystem image, so the clock is still serving the old web UI. The page was driven against a local stand-in for the device's web server that serves these same files — migration display, a wrapping window, an empty list, an incomplete window blocking the save, and a disabled alert keeping its schedule. I am happy to do the filesystem upload and confirm on the device if you would like that before merging.

ktomy:
This one is quite big, I have to look at it more thoroughly to give an opinion. On the other hand this looks like a very low-prio thing, only a very few people would need this. But as the code is there anyway, I will look at it

nishanm:
No argument about the size, and take your time with it.

On the priority though — you're right that for someone with a single clock this is close to useless, and I'd agree with shipping nothing if that were the whole picture. The value is entirely in running **more than one**, which is a fairly common setup precisely because the point of these is to have the number visible wherever you happen to be. Once there are several, they all alert everywhere, all the time, and that turns out to be the problem.

The pattern looks like this:

- one where someone works, useful during weekday working hours
- one in a shared living area, useful in the late afternoon and through the evening
- one where someone sleeps, useful only overnight
- one kept somewhere like a classroom or an office, useful Monday to Friday during those hours and at no other time

Today every one of those alerts on the identical schedule, so the one where someone sleeps is sounding during the day when nobody is in the room, and the one in the living area sounds at 3am. The practical consequence isn't noise, it's **alarm fatigue** — when every device alerts everywhere regardless of whether anyone is there to act, people stop reacting to all of them. An alarm that has been learned as background noise is worse than no alarm, because it is trusted to be there and isn't doing anything.

The workaround today is toggling alerts on and off per device as people move around the house, which nobody does reliably. So in practice people disable alerts on most of their clocks and leave one armed, which means the alert doesn't happen in the room where somebody could actually respond to it.

The last case is the one I'd most like to make work, and it's a bit different from the others: a clock in a shared or institutional space has to be a good citizen or it gets unplugged. "It only makes noise Monday to Friday between these hours" can be the difference between such a device being allowed to stay and not.

None of which is a case for prioritising it — the code is written, it defaults to unchanged behaviour, and it can sit as long as you like. I just didn't want the "few people need this" read to stand unchallenged, because I think the shape is really "few people with one clock need this, and most people with several do".

If the size is the sticking point, I'm happy to split it into the firmware change and the settings page as two PRs, which would make the first one small and reviewable on its own. Say the word.

ktomy:
Yes, in this set up you have a point. How many clocks do you actually have (not hypothetically how many do you need)?

nishanm:
We have 3, one in my office for work hours, one in the living room for when the kids get home and one in the bedroom for over night. We have also provided one to the school which would only sound during school hours. Overkill yes but worth the 40 bucks a pop for convenience.

nishanm:
Two updates while this waits for you, no change to the shape of the feature.

**The day parser was too forgiving.** `parseAlertWindowDays` skipped any character it did not recognise and returned whatever bits were left, and validation only asked whether the result was non-zero. So `"Mon1"` was accepted and stored as Monday, `"17"` became Monday, and the save succeeded while the schedule kept was not the one submitted — on an allow-list that decides when an alarm may sound. The error string already promised digits; the parser just did not hold anyone to it.

It now reports whether the whole value parsed and fills the mask only when it did. Empty, longer than seven characters, a stray character or the same day twice are all refused with a reason. `parseTimeOfDayMinutes` twenty lines below has always been this strict, and two parsers in one file disagreeing about how much they will guess seemed worth closing. The Web UI is unaffected — day values come from one checkbox each, sorted and joined, so it cannot produce either case.

Posted at `/api/save` on a device: `"Mon1"`, `"17"`, `"11"`, an empty list and an eight character list were each refused with the message, and re-reading `config.json` after all of them showed it untouched, so a rejected save writes nothing.

**The settings page has now been exercised on hardware**, which closes the caveat in the description. Driving the real page against the real firmware: Add window is disabled while an alert is off and enables with it, rows add and remove, and a Mon-Fri 02:00–03:00 window saved through the page came back after the restart as exactly `[{"days":"12345","from":"02:00","to":"03:00"}]`. The same save migrated all three legacy silence keys away in one go. I have since uploaded the filesystem to that clock as well, so the page has run from the device itself.

Also noted in the code that removing the superseded key makes the migration one way: firmware old enough to predate windows finds neither key and alerts at any time, which is the loud direction, but a schedule does not survive a downgrade. Seemed better written down than discovered.

Still happy to split the firmware from the UI if that makes it easier to look at.

nishanm:
Not a change request, just a heads-up about two other open PRs that touch the same lines, so nothing surprises you at merge time.

**`tryGetTimezonedTime` is added by both this PR and #190**, with different signatures. This one is `bool tryGetTimezonedTime(tm&)`. #190's is `bool tryGetTimezonedTime(tm&, uint32_t timeoutMs = 0)`, because its caller sits on the render path and must not block when the clock has no time. That is a superset: the callers here compile against it unchanged, so whichever merges second just drops its own declaration and definition.

**The rename in this PR collides with #188.** `sendFaceCycleValidationError` becoming `sendSaveValidationError` is the right call given it now validates more than face cycling, but #188 changes the bounds inside that function and predates the rename. One-line resolution either way.

I have both of those merged together with everything else in a local build; it compiles and has been running on a clock.

nishanm:
**Known item: `tryGetTimezonedTime` overlaps with #190, and the two versions differ in their default timeout.**

This PR adds `ServerManager::tryGetTimezonedTime(tm&)`, implemented as `return getLocalTime(&timeinfo);`. The Arduino core declares `getLocalTime(struct tm*, uint32_t ms = 5000)`, so this version waits up to 5 s for a time sync before reporting failure.

#190 independently adds `tryGetTimezonedTime(tm&, uint32_t timeoutMs = 0)` — the same helper, for the same reason, but defaulting to no wait so the render path never blocks.

Both PRs are individually mergeable against `main`, so this only appears once both land. The signatures are *compatible*: #190's is a superset and this PR's call site compiles against it unchanged. **Whichever merges second, the conflict resolves cleanly and quietly changes the alarm gating here** — from waiting up to 5 s to checking once and giving up (with `ms = 0`, `getLocalTime` runs a single iteration and returns).

For this PR's purposes that is arguably the better default, since an unknown time is meant to resolve toward sounding the alarm and a 5 s block on an alarm decision is undesirable. But it is a behaviour change rather than a no-op, so recording it here rather than letting it land silently.

Suggested resolution when the second of the two merges: keep #190's signature, drop this PR's copy, and pass an explicit timeout here if the waiting behaviour is wanted.

nishanm:
### The settings page and the gating are now hardware-tested

This PR previously carried the caveat that the alert-window settings page had never been exercised on a device. That is now done, on a TC001 with a low reading held in alarm range.

| case | alert windows | result |
|---|---|---|
| no windows configured | `[]` | reading reached the alarm path, **no suppression** — the documented "no windows means alert at any time" no-op |
| window excluding the current time | `Wed 08:00-09:00`, tested at 22:2x | **`outside every alert window, staying quiet`**, 61 times, 0 alarms |

Day parsing, the `{days, from, to}` shape, window evaluation and the save/round-trip all behaved. The settings page was separately exercised off-device against a stub of the clock's web server: a full 76-key config loaded and saved untouched round-trips with **0 keys dropped**.

### One thing this testing turned up, fixed in #190

While the windows themselves were correct, roughly one call in ten logged *"time is not known, ignoring alert windows and alerting anyway"* on a clock whose time was perfectly well synced — and each one leaked an alarm past a window that should have silenced it.

The cause is in the shared `tryGetTimezonedTime` helper, not here: `getLocalTime()` bounds its wait with `while ((millis() - start) <= ms)`, so a zero timeout skips the check entirely whenever the millisecond counter ticks in between. Details and the before/after measurements are on #190, where the fix lives.

Two things worth drawing out for review:

- **This PR's fail-safe is what made the bug visible rather than dangerous.** Treating an unknown time as "sound the alarm" meant the failure surfaced as an alarm that should have been quiet, not as a missed low. A design that resolved the other way would have silently suppressed a low-glucose alarm and produced no symptom at all.
- **Once #190's fix lands, alert windows become genuinely effective.** That is the intent of the feature, but it does mean anyone testing against today's merged behaviour will find it quieter afterwards.

### Note if #190 merges near this one

Both PRs add `ServerManager::tryGetTimezonedTime`. #190's takes an optional timeout and is a superset, so this PR's copy should be dropped when the second of the two merges — see the earlier note on this thread.

nishanm:
Based on your feedback on #177, the same changes are incorporated here: rebased onto main and the long comments trimmed to one or two lines each. The silence-interval migration stays, since those keys shipped in 0.30.0. Builds for ulanzi; no functional change. Note this and #184 touch the same save-validation lambda, so whichever lands second will need a trivial rebase.

nishanm:
One more small change: `tryGetTimezonedTime()` now reads the clock directly instead of through `getLocalTime()`, which blocks for up to five seconds while the clock is unset. It is the same helper #190 carries, so the two no longer differ. Unknown time still means the alarm sounds.

nishanm:
Rebased onto main after #184, #188 and #185 landed. The seed config keeps your default melodies alongside the new alert-window keys. Builds for ulanzi; no functional change.

nishanm:
Correction to my last comment: you had already merged main into this branch and removed the migration before my push, so my rebase was rejected and nothing of mine landed. The branch is exactly as you left it.

ktomy:
@nishanm please don't commit for now, I am working on this PR

PR #184, opened 2026-09-07
Make the alarm repeat interval configurable

nishanm:
### What changed

The pause between alarm repeats was fixed at 5 minutes (`ALARM_REPEAT_INTERVAL_SECONDS`). It is now chosen in the web UI — 1, 2 or 5 minutes — and **defaults to the 5 minutes it has always been**, so an existing clock behaves identically until somebody changes it. The choice is deliberately only in the more insistent direction; see the discussion below.

Intensive mode is untouched and still overrides the setting with its 2 second repeat.

### Why

Five minutes is a long time to wait when a reading is heading the wrong way, and too often when it is sitting just over the line and already being dealt with. Which of those is true depends on the person and the clock, so it is not something a single constant can get right.

### Notes for review

- Validated in two places: on load, defaulting to 300 if the stored value is not one the dropdown offers, in the style of the existing `isValidFaceCycleInterval`; and at `/api/save` before anything is written, so a posted value that is not supported is refused rather than accepted and quietly corrected on the next boot. `face_cycle_interval_seconds`, the other enumerated setting in that handler, was already checked there.
- `isValidAlarmRepeatInterval` moved from an anonymous namespace to a public static so the endpoint asks `SettingsManager` which values it accepts instead of restating the list. The error helper in `ServerManager` is now `sendSaveValidationError`, since it no longer validates only face cycling.
- The "How alarms work" panel said "pause for 5 minutes"; it now points at the setting.
- Independent of #183 — the two touch the same alarm card but at different anchors, and merge cleanly in either order.

### How it was tested

Builds clean on `ulanzi_debug`. The web UI was driven against a local stand-in for the device's web server: a config without the key loads the 300 default, and a changed selection reaches the saved configuration as `alarm_repeat_interval_seconds`.

**Not yet run on real hardware.** I have live units and did not want to reflash one before you had seen the approach. Say the word and I will do a hardware pass and report back before you merge.

### DIFF
--- a/data/config_initial.json
+++ b/data/config_initial.json
@@ -44,6 +44,7 @@
     "alarm_low_melody": "",
     "alarm_urgent_low_melody": "",
     "alarm_intensive_mode": false,
+    "alarm_repeat_interval_seconds": 300,
     "librelinkup_email": "",
     "librelinkup_password": "",
     "librelinkup_region": "EU",
--- a/data/index.html
+++ b/data/index.html
@@ -657,7 +657,19 @@
                             </div>
                         </div>
                         <div class="row mt-3">
-                            <div class="col-12">
+                            <div class="col-sm-4">
+                                <label for="alarm_repeat_interval_seconds" class="form-label">Repeat every</label>
+                                <select class="form-select" id="alarm_repeat_interval_seconds" required>
+                                    <option value="60">1 minute</option>
+                                    <option value="120">2 minutes</option>
+                                    <option value="300">5 minutes</option>
+                                </select>
+                                <div class="form-text">
+                                    How long the alarm waits before sounding again while the reading stays
+                                    out of range, when nobody has snoozed it. Intensive mode ignores this.
+                                </div>
+                            </div>
+                            <div class="col-sm-8 align-content-center">
                                 <div class="form-check">
                                     <input class="form-check-input" type="checkbox" id="alarm_intensive_mode">
                                     <label class="form-check-label" for="alarm_intensive_mode">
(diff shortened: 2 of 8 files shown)


ktomy:
But why not just snooze the alarm if it sounds to often? agree with 1 and 5 minutes, but having bigger intervals... I don't know...

nishanm:
Fair on both counts, and I've cut the long end — the dropdown is now **1, 2 and 5 minutes only**, still defaulting to 5. Pushed.

On why not just snooze: snooze is the right tool for "I've seen this, be quiet for a while", and it covers that better than a setting could, because it's a decision somebody makes with the reading in front of them. But snooze can only ever make the clock *quieter*, and it needs someone to press the button. What it can't do is make the clock more insistent when nobody has acknowledged anything yet — and 5 minutes is a long wait for the second beep if a reading is heading the wrong way and the person who needs to act is in another room.

So the useful half of this is the short end, which is exactly the half snooze cannot reach. The long end you were uneasy about is the half snooze already does properly, and on an alarm people rely on I would rather not ship a setting whose main effect is to make it less likely to be heard. Dropping 10, 15 and 30 costs nothing real and removes that.

Happy to go further and make it 1 and 5 only if you would prefer the smaller set — 2 is in there because it is a noticeably different feel from 1 without being nearly as insistent, but I do not feel strongly about it.

ktomy:
LGTM, I'll test on the device and will merge

nishanm:
One addition before you test this, since it was a gap rather than a preference.

The interval was validated on load but not at `/api/save`, so a posted value that is not one of the three was written to `config.json`, served back to the settings page, and only quietly became 300 on the next boot — the file saying one thing and the clock doing another. `face_cycle_interval_seconds`, the other enumerated setting in the same handler, has always been checked before the write; this one just was not.

It is now checked at the endpoint too. `isValidAlarmRepeatInterval` moved from an anonymous namespace to a public static so the endpoint asks `SettingsManager` which values it accepts rather than restating the list — the face cycle check above already spells its set out twice, in the message and the comparison, and this avoids adding a third copy. The error helper is renamed `sendSaveValidationError` since it no longer validates only face cycling.

Verified on a device: 45, 0 and the string `"300"` are each refused with the message, 60/120/300 accepted, and `config.json` is untouched after a rejection.

No change to the dropdown or the default — still 1, 2 and 5 minutes, still defaulting to the 5 it has always been.

nishanm:
Based on your feedback on #177, the same changes are incorporated here: rebased onto main, and the comments trimmed to say what rather than why. No functional change.

PR #185, opened 2026-09-07
Add a picker of ready made alert sounds

nishanm:
### What changed

Adds an alert sound picker so each high, low, and urgent-low alarm can use a recognizable sound without entering RTTTL by hand. The six presets are Double beep, Triple beep, Two tone siren, Urgent pulse, Soft ping, and Long tone. Each alarm also has a Default option and an editable custom RTTTL field.

### Configuration and preview behavior

- Selecting a preset or Default fills the melody field with its RTTTL string. Editing the field updates the picker to the matching option or Custom.
- Save stores the actual RTTTL string, including for Default. The three original alarm melodies are supplied by `data/config_initial.json` and the Default options in the UI; the firmware alarm melody constants and fallbacks are removed.
- Try previews the current field without saving, including when the alarm is disabled. For example, after saving Soft ping, selecting Default and clicking Try plays Default immediately.
- Try sends `{ "rtttl": "..." }` to `/api/alarm`. This replaces both `/api/alarm/custom` and the former `/api/alarm` request accepting `alarmType`.
- Empty or invalid melody fields block preview. Enabled alarms require a nonempty melody that passes the existing RTTTL validation before saving.

Options remain plain HTML `<option>` elements, consistent with the existing snooze selectors. Alarm thresholds, scheduling, and snoozing logic are unchanged.

The final change includes firmware and filesystem changes, so both images must be deployed. No migration for empty melody settings is included: this project uses a full reflash, including configuration, for version upgrades.

### Validation

- The contributor reports auditioning all six presets on a TC001 using the earlier custom-preview endpoint. Rising and falling chirps were removed, and the siren was retuned after listening on hardware.
- For commit `6f80656`, `scripts/build.sh --all` passed for firmware and LittleFS. JavaScript syntax and whitespace checks also passed.
- Focused JavaScript checks passed for all three alarms: UI/config default consistency, preset/default serialization and reload, custom selection, disabled controls, and empty-field rejection. Preview checks covered all 21 preset/default combinations and the saved-Soft-ping to unsaved-Default case; the final requests use `/api/alarm`.
- Still pending for the final implementation: complete browser/device save-and-reload checks, Try with disabled alarms, actual triggered-alarm playback including intensive mode, and mobile layout verification. The final firmware has not been flashed during this review.

### DIFF
--- a/data/index.html
+++ b/data/index.html
@@ -514,9 +514,21 @@
                             </div>
                         </div>
                         <div class="row mt-2">
-                            <div class="col-12">
-                                <label for="alarm_high_melody" class="form-label">Custom melody (RTTTL)
-                                    (optional)</label>
+                            <div class="col-sm-4">
+                                <label for="alarm_high_melody_preset" class="form-label">Alert sound</label>
+                                <select class="form-select" id="alarm_high_melody_preset" disabled>
+                                    <option value="">Built in default</option>
+                                    <option value="doublebeep:d=8,o=6,b=180:c,p,c">Double beep</option>
+                                    <option value="triplebeep:d=16,o=6,b=200:c,p,c,p,c">Triple beep</option>
+                                    <option value="siren:d=4,o=5,b=100:a,d6,a,d6">Two tone siren</option>
+                                    <option value="urgent:d=32,o=7,b=220:c,p,c,p,c,p,c,p,c,p,c">Urgent pulse</option>
+                                    <option value="ping:d=4,o=6,b=140:8e,16p,8c">Soft ping</option>
+                                    <option value="longtone:d=1,o=5,b=90:a">Long tone</option>
+                                    <option value="custom">Custom</option>
+                                </select>
+                            </div>
+                            <div class="col-sm-8">
+                                <label for="alarm_high_melody" class="form-label">Melody (RTTTL)</label>
                                 <input type="text" class="form-control" id="alarm_high_melody"
                                     placeholder="high:d=4,o=5,b=125:4e7,p,4e7" disabled />
                                 <div class="form-text">Leave empty to use the default high alert sound.</div>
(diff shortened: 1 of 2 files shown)


ktomy:
Hi, is there a specific reason you keep templates in JS array instead of having them directly in the `<option/>` in HTML? I think it would made the code simpler and would mean less changes

nishanm:
No good reason, and you're right — moved. The options are now plain `<option>` elements in `index.html`, and the JS array and the function that populated the selects are both gone. Pushed.

My reason was avoiding repeating eight options three times, once per alert level. But I went and looked at the file, and the snooze dropdown is already spelled out in full three times — so the array wasn't consistent with the rest of the page, it was just my habit. Following the existing pattern wins.

It came out better than a straight move, too. `syncMelodyPreset` used to search the array to decide whether a typed melody was a known preset; now that the option values *are* the melodies, selecting by value is the whole lookup:

```js
preset.val(($(`#alarm_${alarmType}_melody`).val() || '').trim());
if (preset.val() === null) {
    preset.val('custom');
}
```

A melody that isn't in the list simply matches no option, which is exactly what "Custom" is for. Net effect on the JS is 34 added lines instead of 50, with no module level state at all.

Re-checked in a browser afterwards: picking a preset fills the field and it validates, typing something of your own selects "Custom", clearing the field selects "Built in default", and pasting a melody that matches a preset selects that preset.

Still worth repeating from the PR description: **I have not actually heard these on the buzzer yet.** They're valid RTTTL and use only characters the existing validator accepts, but whether they sound good on that speaker is untested, and for a sound feature that's the part that matters. Happy to flash a device and report back, and equally happy to swap any of them for something you'd rather ship.

ktomy:
> I have not actually heard these on the buzzer

Wait, this is important. Before coming up with the sounds which are on the device today I spent quite a few hours testing various tones. The buzzes is not great on the device and a lot of things which sounded ok in the emulator did not sound well on the device. Please test things on the device and come up with those you think make sense. Thanks

nishanm:
Update after actually hearing these on hardware — which I should have done before opening this.

**Two presets are gone.** "Rising chirp" and "Falling chirp" were unpleasant on the real buzzer in a way that isn't obvious from the RTTTL. Six left: double beep, triple beep, two tone siren, urgent pulse, soft ping, long tone.

**The siren is softer.** The original was `d=8` eighth notes at 120bpm alternating A5/E6, which came out shrill and frantic. It's now `d=4` at 100bpm with the top note dropped to D6 — same character, noticeably less piercing.

So the "these have never been heard on a buzzer" caveat from the description is now closed: every remaining preset has been played on a TC001 through the device's own `/api/alarm/custom` endpoint, and the two that didn't survive have been cut.

One thing worth flagging, because it affects how loud any of these are: **`setVolume()` does nothing on ESP32 in the vendored MelodyPlayer.** In `lib/MelodyPlayer/melody_player.cpp` the `ledcWrite(pwmChannel, volume)` call that applies it sits behind `if (false)`, and the blocking play path never applies volume at all. I found it while trying to make an early-low warning quieter, and measured it: six volume levels from 250 down to 0 were completely indistinguishable, then clearly different once the write was enabled. That's in a separate PR since it touches `lib/` and is a bigger question than this one.

It matters here because it changes what this PR is for. With volume working, somebody who finds a preset too loud can turn that alert down instead of needing a gentler melody — so this list can stay small and characterful rather than trying to cover every loudness preference. If you'd rather I re-tune any of the six, or drop more of them, happy to.

nishanm:
Corrected the description, which was still describing the first version of this PR — it said eight presets, referred to a `MELODY_PRESETS` constant that no longer exists after the move to `<option>` elements, and claimed all eight had been validated. It now says six, and matches the diff.

No code change in this push; the six hardware-auditioned presets are as reported above.

nishanm:
### Heard on hardware

This PR previously carried the caveat that the presets had never been played on a real buzzer, which for a sound feature is the only verification that counts.

All six presets have now been played on a TC001 via `/api/alarm/custom`: **Double beep, Triple beep, Two tone siren, Urgent pulse, Soft ping and Long tone.** Every one produced audible output and they are distinguishable from one another.

That closes the outstanding caveat on this PR. It remains `data/` only, with no firmware change.

nishanm:
Based on your feedback on #177, the same changes are incorporated here: rebased onto main, and the one long comment trimmed. Still data/ only, no functional change.

PR #187, opened 2026-09-08
Make the alarm volume configurable, and fix the volume code it depends on

nishanm:
Approach proposed in #186, opened alongside it so the shape is visible — I would rather agree the approach there than have you review a diff into `lib/` cold. Happy to rework or close this if you would rather it were done differently.

### What changed

Each alert can be set to **High, Medium or Low**, chosen separately for high, low and urgent low. Everything defaults to High, which is the value that reproduces exactly what alerts sound like today, so an existing clock is unchanged until somebody chooses otherwise.

The case that prompted it: an early low warning wants to be a gentle nudge, while urgent low wants full volume. Today every alert is equally loud, which pushes people toward switching the gentler ones off entirely.

### The part that needs your steer: a fix in `lib/`

**This could not work without it.** In `lib/MelodyPlayer/melody_player.cpp` the write that applies volume to the hardware sits behind `if (false)`:

```cpp
ledcWriteTone(player->pwmChannel, computedNote.frequency);
if (false)
    ledcWrite(player->pwmChannel, player->volume);
```

and the blocking play path never applies volume at all. `setVolume()` has been storing a value nothing ever wrote. That is upstream MelodyPlayer's code, not this project's — it came in with the vendoring commit.

Two lines are enabled, plus a comment recording the ordering constraint: `ledcWriteTone()` sets duty to 50% itself, so the volume write has to come *after* it or it is overwritten.

I am happy to report this upstream to MelodyPlayer as well — it affects every project using the library.

### Notes for review

- `DEFAULT_ALARM_VOLUME` and the three levels are defined once in `globals.h`. MelodyPlayer halves the value it is given, so 250/125/60 become PWM duty 125/62/30 of 255. 125 is roughly half duty, the loudest a square wave gets, which is why High is the no-change default.
- Volume is set immediately before each melody, deliberately every time, so one alert's volume cannot leak into the next.
- A value that is not one of the three falls back to the default rather than being guessed at.
- Try plays at the volume that alert will actually use, so it tests the real thing rather than a fixed loudness.
- Independent of #183, #184 and #185 — all four merge cleanly in any order.

### How it was tested

Measured on a TC001, not reasoned about. **Before the fix**, six volumes from 250 down to 0 were indistinguishable, including 0, which should have been silent — that null result is what sent me to read the library. **After it**, the levels are clearly different and 0 is silent.

Then end to end through the real settings path, not just the test endpoint: the volume was written to the device config, survived a reboot, and the alert played at the configured level. Checked on both the high and urgent low paths, since each alert carries its own volume and I wanted the per-alert wiring exercised rather than assumed.

Builds clean on `ulanzi_debug`. The settings page itself has only been exercised against a local stand-in for the device web server, since I flashed firmware only to keep the device's configuration intact.


nishanm:
Based on your feedback on #177, the same changes are incorporated here: rebased onto main, comments trimmed, and the one-argument `playRTTTLString` overload removed since nothing calls it any more. Also fixed a nesting slip in `index.html`: the urgent-low volume row was inside the melody row, unlike high and low. Builds for ulanzi; no functional change.

nishanm:
Rebased onto main after #185 reworked the alarm endpoint. The volume now rides along with the melody in the `/api/alarm` payload (optional, defaults to the level alerts always played at), and the per-alarm resolver this PR used to add is gone with the old endpoint. Builds for ulanzi; same feature as before.

ktomy:
I will test it on the device later today and will let you know what I think,

ktomy:
Rebased onto current `main` (`d4ba07c`) and resolved conflicts with the settings-page rewrite.

Ported the per-alert High/Medium/Low volume selector, numeric handling, and validation into `web/src/js`. “Try on clock” sends the selected volume with the melody. Kept the obsolete `data/index.html` and `data/script.js` files deleted, while retaining the firmware and MelodyPlayer volume fixes.

PR #181, opened 2026-09-06
Add Unicorn mascot clock face

JuanMiste:
New 6th clockface: a unicorn sprite carries the alarm state through its mane color (rainbow when normal, flattened to the warning/urgent/stale color otherwise), with the glucose value shown beside it.

My daughter asked for a unicorn face, and I built it with Claude's help, including tracing the sprite pixel-by-pixel against a reference photo.

Contribution built by Claude AI

PR #188, opened 2026-09-09
Derive clock face id bounds from one constant instead of a literal 6

nishanm:
Face ids are bounds-checked in three places, and all three used a literal `6`:

- the `/api/save` validator, with `bool selectedFaces[6]` and `faceId >= 6`
- the face-cycle list when loading a config, with `faceAlreadyAdded[6]`
- the default-face fallback, `default_clockface < 6`

That is correct today with six faces. It stops being correct the moment a seventh is added: the firmware renders the new face happily, but saving settings fails with *"Face selections must use IDs from 0 to 5"*, and the config load path indexes one past the end of `faceAlreadyAdded`. Anyone adding a face has to find all four sites to avoid that, and nothing points them there.

All of them now come from a single `CLOCK_FACE_COUNT` in `globals.h`, and `BGDisplayManager` logs a mismatch at startup if the number of registered faces ever drifts from the constant.

**No behaviour change with the current six faces.** The validator's message becomes "Face selections must use valid clock face IDs" rather than naming 0 to 5, so it does not go stale next time.

I found this while adding a clock face in a separate branch, but it is useful on its own and there are other open PRs that add faces (#170, #181) which would each hit it, so I have kept it separate rather than burying it in a feature.

Built for `ulanzi_debug`. Not separately exercised on hardware beyond building and running, since with six faces every bound evaluates exactly as before.

### DIFF
--- a/data/index.html
+++ b/data/index.html
@@ -731,12 +731,6 @@
                                     face</label>
                                 <select class="form-select" id="default_clock_face" required
                                     aria-describedby="default_clock_face_cycle_help">
-                                    <option value="0">Simple</option>
-                                    <option value="1">Full glucose graph</option>
-                                    <option value="2">Glucose graph and value</option>
-                                    <option value="3">Big text</option>
-                                    <option value="4">Value and delta</option>
-                                    <option value="5">Current time and BG value</option>
                                 </select>
                                 <div class="invalid-feedback">
                                     Please select default clock face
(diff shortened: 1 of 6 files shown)


nishanm:
A note for whenever this is merged, since it touches a line #183 also changes.

#183 renames `sendFaceCycleValidationError` to `sendSaveValidationError`, because it reuses the same helper for validating alert windows. This PR changes the bounds inside that function but keeps the old name, so the two conflict on those lines.

The resolution is mechanical — take the `CLOCK_FACE_COUNT` bounds from here under whichever name is current — and I have verified it by merging both into a local build, which compiles and runs on a device.

Worth saying plainly that this PR is not urgent on its own: with six faces every bound evaluates exactly as it does today, so nothing is broken right now. It only matters the moment a seventh face exists, and there are two open PRs adding faces (#170, #181). If it is easier to take those first and fold this in afterwards, that works equally well.

PR #190, opened 2026-09-09
Change the face and brightness on a daily schedule

nishanm:
Rows of time, face and brightness under Device settings. From each time the clock shows that face at that brightness until the next row, and the last row of the day runs overnight. Two rows make a bedside clock: Big text on an automatic mode in the morning, Simple (dark) at brightness 1 in the evening.

```json
"face_schedule_enabled": true,
"face_schedule": [
    { "time": "07:00", "face": 3, "brightness": 101 },
    { "time": "22:00", "face": 7, "brightness": 1 }
]
```

`brightness` is the same value the brightness dropdown uses: 1 to 10 for a manual level, 100 and 101 for the two automatic modes.

### How a row is applied

The way the Web UI or the buttons would apply it: the brightness settings change in memory, so the automatic modes carry on from there, and the face is switched. There is no restore logic; the next row is the restore.

- On boot or after a time sync, the latest row whose time has passed applies. A clock that does not know the time stays on its default face.
- The buttons still change the face between rows.
- The schedule and automatic cycling are exclusive: the Web UI lets only one be switched on, and cycling wins if a config somehow has both.
- Rows are validated in the Web UI (a time on every row, no duplicates, at most 8). The loader only drops rows it could not apply.

The `"HH:MM"` helpers from the alert windows are shared rather than copied, so they move out of that file's anonymous namespace.

Builds for ulanzi. Verified off-device against a stub of the clock's web server: a config without the keys loads with the schedule off; enabling it greys out cycling and vice versa; a save posts the rows as above; a config with rows renders them, and a row with an unknown brightness falls back to Auto: balanced.


nishanm:
Two updates.

**Squashed to a single commit.** The branch had picked up ten commits while the design changed shape — an earlier version added two purpose-drawn low-light clock faces and a thin 4x8 font, and that was later dropped in favour of reusing the faces that already exist. The net diff never included any of it, but the history made it look like the PR added and then deleted a font. The tree is byte-identical to what was here before the squash.

**The settings page has now run on the device.** The description says it was verified against a stub of the clock's web server but not on hardware; that is no longer true. I flashed both the firmware and the filesystem image to a TC001 and the settings page it serves has the night mode block rendering correctly. Config survived by being restored through the AP window afterwards.

The one claim still argued from code rather than observed is the unsynced-clock case — that no NTP must mean "not night", so a clock that has never got the time cannot dim itself and stay dim. I could not induce that on a healthy device.

**One thing to be aware of when merging.** This adds `ServerManager::tryGetTimezonedTime`, and so does #183 — with a different signature. #183's takes only the `tm&`; this one takes an optional timeout defaulting to zero so callers on the render path never block. The two collide, but the version here is a superset and #183's callers compile against it unchanged, so whichever merges second only needs the other's declaration dropped. I have both merged together in a local build and it builds and runs.

nishanm:
**Known item: `tryGetTimezonedTime` overlaps with #183, and the two versions differ in their default timeout.**

This PR adds `ServerManager::tryGetTimezonedTime(tm&, uint32_t timeoutMs = 0)`. #183 independently adds `tryGetTimezonedTime(tm&)` with no timeout parameter, whose body is `return getLocalTime(&timeinfo);` — and the Arduino core declares `getLocalTime(struct tm*, uint32_t ms = 5000)`, so #183's version waits up to 5 s for a time sync.

Both PRs are individually mergeable against `main`, so this only appears once both land. What makes it worth flagging is that the signatures are *compatible*: this PR's is a superset, and #183's call site compiles against it unchanged. **Whichever merges second, the conflict resolves cleanly and quietly changes #183's behaviour** — its alarm gating would go from waiting up to 5 s to checking once and giving up (with `ms = 0`, `getLocalTime` runs one iteration and returns).

For an alarm path, failing fast is arguably the better default: an unknown time means "sound the alarm", so the fast path errs toward being noticed and avoids blocking an alarm decision for five seconds. But it is a behaviour change rather than a no-op, and it should be a decision rather than a merge artifact.

Suggested resolution when the second of the two merges: keep this PR's signature, drop #183's copy, and let #183's call site pass an explicit timeout if it wants the waiting behaviour back.

nishanm:
### Merge order for the nine open PRs, and the collisions between them

GitHub reports all nine individually `MERGEABLE` against `main`, which is misleading — several conflict with **each other**, not with `main`. I merged the whole set locally in the order below, resolved every conflict by hand, and the result builds. So this is a tested order with known resolutions, **not** a claim that they auto-merge.

**Order:** #177 → #178 → #188 → #190 → #191 → #184 → #187 → #185 → #183

Anything can be merged alone; the order only matters when taking several.

**Where they actually collide**

- **#178 on #177** — stacked, so trivial in this order.
- **#191 on #177/#178** — the largest one, 8 files. #177 replaces `BG_COLOR_OLD` with `getDataOldColor()` and #178 adds the early-stale tier, while #191 refactors the same two functions to take an explicit colour. ⚠️ **Two hunks here merge with no conflict marker and are wrong afterwards**: `showReading()` keeps `BG_COLOR_OLD` instead of the `getDataOldColor()` / early-stale precedence. Worth grepping for `BG_COLOR_OLD` after merging rather than trusting the conflict list.
- **#191 × #177 in `globals.h` / `enums.h`** — #177 defines the `COLOR_*` macros from a `DISPLAY_COLOR` enum that already contains `MAGENTA`; #191 adds `COLOR_MAGENTA` as a literal. Take #177's enum-derived version and define `NIGHT_VALUE_COLOR` from `DISPLAY_COLOR`.
- **#178 × #190 in `data/index.html`** — the Early Stale block and the Night mode block are both appended at the same anchor and each completes the *same* half-open `<input>`, so the union needs its own copy of the surrounding scaffolding rather than a naive concatenation.
- **#184 × #188** — #184 renames `sendFaceCycleValidationError` to `sendSaveValidationError`. Keep the new name with #188's `CLOCK_FACE_COUNT` bounds. #183 carries the same rename.
- **#184 × #187** — both add a helper in the same place in `SettingsManager.cpp`. Keep both.
- **#183 × #190** — both add `ServerManager::tryGetTimezonedTime`. #190's takes an optional timeout and is a superset; whichever merges second should drop its own copy. The default differs and that changes behaviour — see the note already on this PR and on #183.

Happy to rebase any of these myself once earlier ones land, or to split #191 off #190/#188 if you would rather take them separately.

nishanm:
Two updates: a defect fixed, and the fail-safe this PR promised to test is now tested.

### Fixed: `night_face = -1` did not do what it said

The sentinel is documented in `Settings.h` as *"-1 keeps the current face and only caps brightness"* and the dropdown option reads "Keep my normal face". But it fell into the same branch as an out-of-range id and switched to `default_clockface`. Anyone on a non-default face — because face cycling moved them, or they pressed a button — was pulled to the default when the window opened.

The fix is slightly larger than it first looks, because the bug is asymmetric: entry ignored the sentinel, while exit **unconditionally** restored `faceBeforeNight`. Fixing only the entry would have left night-end dragging users off a face they had deliberately chosen during the window. So the swap is now tracked:

- `-1` leaves the face alone entirely and only caps brightness;
- an out-of-range id still falls back to `default_clockface`, since that is a broken config rather than a request;
- exit only undoes a swap this code actually made.

Also removed two stray blank lines this PR had introduced, which takes `BGDisplayManager.cpp` to zero clang-format violations.

### Verified on hardware: an unsynced clock is not treated as night

This PR previously said the fail-safe was argued from code and never observed. It has now been observed on a TC001, with the window set to `00:00`–`23:59` so that night **must** activate if the clock knows the time at all.

| condition | time known | night mode activated |
|---|---|---|
| normal firmware | yes, `17:28:38` | **yes** — `Night mode on, switching to face 6` |
| clock forced unsynced | no, repeated `Failed to obtain time` | **no — 0 activations in 200 s** |

Both observations are from the same device and the same config in one sitting, differing only in whether the clock could read the time.

The control that matters for the negative case: the render loop was demonstrably still running throughout — 33 reading/tick log events in the same capture — so the absence of night mode is the fail-safe working, not the firmware being wedged.

**Two earlier attempts at this test failed to induce the condition, which is worth recording because both look like they should work:**

1. Repointing NTP at an unroutable address and resetting over USB. The ESP32's RTC retains time across a soft reset, so the clock never needed NTP and reported the correct time regardless.
2. Zeroing the clock with `settimeofday()` before `configTime()`. Time still resolved within the 5 s `getLocalTime()` window — most likely because NTP is intercepted on the network this bench sits on, so an unroutable server still got an answer.

Only forcing `initTimeIfNeeded()` to leave the clock unset, independent of the network, actually produced the "time is unknown" state the fail-safe guards. A reviewer reproducing this should expect the first two approaches to give a false pass.

nishanm:
### Defect found on hardware and fixed: a zero timeout could skip the time check entirely

Following up on the note above about `tryGetTimezonedTime`'s default. It is not only a difference of opinion with #183 — the zero default is a bug in its own right, and it has now been reproduced on a device.

`getLocalTime()` bounds its wait like this:

```c
uint32_t start = millis();
while ((millis() - start) <= ms) { ... }
```

With `ms == 0`, the loop body is skipped entirely whenever the millisecond counter ticks between those two reads. It then returns false **without ever looking at the clock**, so `tryGetTimezonedTime()` reports an unknown time on a device that is perfectly well synced.

Measured on a TC001 running this branch merged with #183, with a low reading held in alarm range and an alert window that excluded the current time:

| `timeoutMs` | "time is not known" | alarms that should have been suppressed | correct suppressions |
|---|---|---|---|
| `0` (this PR's default) | 7 | **7** | 61 |
| non-zero | 0 | **0** | 61 |

Same device, same config, same reading, same duration — only the timeout differed.

**The fix keeps the non-blocking intent rather than raising the default.** A zero timeout now reads the clock directly and checks `tm_year`, which is exactly what `getLocalTime()` does on its first iteration, minus the racy loop guard. Non-zero timeouts still go through `getLocalTime()` unchanged.

Re-verified after the fix on the same hardware and config: **0 unknown-time events, 0 leaked alarms, 61 suppressions.**

Worth stating the direction of the old failure: it made an alert window leak, so an alarm sounded when a schedule should have silenced it. That is the safe direction, and it is why this surfaced as noise rather than as a missed low. It also means the fix makes alert windows genuinely quieter, which is their purpose but is a real behaviour change for anyone relying on the buggy behaviour.

### Also in this branch since the last push

- **`night_face = -1` now keeps the current face**, as `Settings.h` and the "Keep my normal face" label always claimed. It previously fell into the same branch as an out-of-range id and switched to `default_clockface`. The bug was asymmetric — entry ignored the sentinel while exit unconditionally restored the pre-night face — so the swap is now tracked and exit only undoes a swap this code actually made.
- Removed two stray blank lines this PR had introduced; `BGDisplayManager.cpp` is now clean under the project's clang-format.
- The **unsynced-clock fail-safe is hardware-verified** (see the earlier comment): with the window covering the whole day, a clock that cannot read the time produced **0 night activations in 200 s**, with 33 reading/tick events in the same capture proving the loop was still running.

nishanm:
Based on your feedback on #177, the same changes are incorporated here: rebased onto main (this PR had drifted to CONFLICTING after the merge), comments trimmed, and the five `night_*` keys added to `config_initial.json` with night mode off, matching how the other features seed theirs. Also removed one stray closing `</div>` after the face-cycle checkboxes. Builds for ulanzi; no functional change.

nishanm:
One more small change: dropped the unused timeout parameter from `tryGetTimezonedTime()`, so it is now the identical helper #183 carries and the two PRs no longer differ there.

nishanm:
Also shortened the night mode help text on the settings page to three lines, and it now mentions that face cycling pauses while night mode is active.

nishanm:
Rebased onto main after #184, #188 and #185 landed. The night face picker is now filled from the shared face dictionary #188 introduced, so it has no copy of the face list of its own. Builds for ulanzi; no functional change.

ktomy:
This comment is for this PR and for #191 as I think they serve more or less same purpose.

The idea of showing different faces on schedule is great. But I would not tie it to day/night. Why not just have a schedule? e.g. 3:00 Face1, 8:00 Face3, 18:00 Face6, etc
Brightness is a different problem. But I totally understand why in some cases you may consider the current brightness modes do not suit your environment.
Then my proposal is:
In the schedule the user selects face and brightness.
e.g. 
07:30 - Face 1 - Auto (balanced)
12:00 - Face 2 - 10
18:00 - Face 3 - Auto (balanced)
20:00 - Face 4 - Auto (dark rooms)
23:00 - Face 5 - 1

If the brightness modes are not enough and you want a new brightness mode (e.g. for even darker rooms), just create a new brightness mode, there are 2 auto modes already, I don't mind having 3 or 4

If you want a new face e.g. one with the arrow colored in the BG color, just create new face :) 

Thank you for your understanding

nishanm:
A schedule is better than night mode. My own setup is two rows of it (07:00 Big text / auto dark rooms, 22:00 Simple (dark) / 1), so I'll rework this PR into that. The shape I'd build:

- Rows of time, face, brightness. Each row sets both, applied at that time exactly as if picked in the UI. No restore logic: the next row is the restore.
- On boot or after a time sync, the most recent row that has passed applies, wrapping midnight. No time known means no row applies, so the fail-safe stays.
- With face cycling on, a row's brightness still applies and the face is left to cycling, the same way the default-face setting is disabled while cycling. Buttons keep working between rows.
- Empty schedule is today's behaviour. Validation in the Web UI only. Any cap you want on the row count?

Brightness per row is the existing list (manual levels plus the two auto modes); a darker auto mode can come on its own if it turns out to be needed. Sound right before I build it?

ktomy:
Sorry, I missed your question and assumed you started working on it.
I think this feature is incompatible with cycling, so they should be mutually exclusive

nishanm:
Mutually exclusive works, and it removes the only fiddly part. In the UI, turning on cycling greys out the schedule card and vice versa, like the default-face select today. I'll cap it at 8 rows unless you want another number. Starting on it.

nishanm:
Rebuilt as the schedule, one commit on top of 0.31. Rows of time, face and brightness under Device settings; each row is applied the way the UI or the buttons would apply it, the last row runs overnight, and the schedule and cycling are exclusive in the UI with cycling winning if a config has both. Capped at 8 rows. The night mode keys are gone. Description updated. Builds for ulanzi.

ktomy:
There is one more thing, if there is one schedule, it will be ignored, e.g. I schedule BigText for 22:00, at 22:10 I switch to a different face using buttons. It will not return back to BigText at 22:00 next day. So I think we should choose one of 2 options: either make it reset to the selected face at the selected time even with one schedule, either allow a minimum of 2 schedules in Web UI.

nishanm:
Went with the first option: a row now re-applies at its time every day, even when it's the only row, so switching faces with the buttons at 22:10 returns to BigText at 22:00 the next day. One small commit on top of yours. Builds for ulanzi.

PR #191, opened 2026-09-09
Add a Simple (dark) face

nishanm:
Simple (dark), face 7: Simple without age bars, with the number in one calm colour (white by default, or magenta or blue) and the trend arrow in the glucose range colour.
- Urgent lows and highs turn the number red, so an urgent reading with no trend arrow is still obvious.
- Stale data uses the data-old colour like every other face.
- The colour lives in a Simple (dark) drawer on the Clock faces card (the mechanism from #178), and the page refuses a number colour equal to the data-old colour.

The earlier five per-band colours are gone. Depends on #178 (→ #195 → #192); only the last commit is new.


nishanm:
### Every colour state is now observed on hardware

The PR body previously qualified its hardware claim, because the face had only been seen on a build that merged this branch with several other PRs. The full colour matrix has now been driven deliberately and read off the panel at `night_brightness_level` 1:

| state | expected | observed |
|---|---|---|
| in range | value in the chosen colour, **green** arrow | ✅ |
| warning low | **yellow** arrow | ✅ |
| urgent low | **red** arrow | ✅ |
| warning high | **yellow** arrow | ✅ |
| urgent high | **red** arrow | ✅ |
| older than the too-old threshold | **whole face cyan**, arrow included | ✅ |
| no readings at all | `showNoData()` in the chosen colour, **not gray** | ✅ |

The value stayed in its configured colour across every band, which is the inversion this face exists for: the reading never lights red or green, and the glucose band is carried by the five-pixel arrow instead.

The `showNoData()` case is worth calling out separately — on the base face that screen draws in gray, which is exactly the colour this panel cannot render at minimum brightness. Here it draws in the chosen value colour, so "no data" stays visible.

**Colour choice:** having compared white, magenta and blue on the panel in a dark room, magenta was preferred. The setting keeps all three, and the shipped default remains **white** — the most legible of the three for anyone who has not looked at their own panel yet.

### Stacking

Still stacked on #190 and #188, and #190 has since gained two fixes (the `night_face = -1` behaviour and a `tryGetTimezonedTime` race). This branch has been rebuilt on top of the current #190 and rebuilds clean.

nishanm:
Based on your feedback on #177, the same changes are incorporated here: rebased onto #190 and main (this had gone CONFLICTING after the merge), the stale-reading colour on the shared text base now goes through `getDataOldColor()`, the night value colour is a `DISPLAY_COLOR` read with `displayColorFromString()` instead of a second enum with its own parser, `night_value_color` is seeded in `config_initial.json`, and the comments are trimmed. One comment claimed the face was kept out of the default-face and cycling lists; the markup was right and the comment was stale, so it is fixed. Builds for ulanzi; no functional change.

nishanm:
Rebased onto #190 and main after #184, #188 and #185 landed. The #188 commit this PR carried is gone since it is merged, and the face is now added to the shared face dictionary rather than to the markup. Builds for ulanzi; no functional change.

nishanm:
This is the "arrow in the BG colour" face you describe, so it stays as it is. Once #190 becomes a schedule it's just a face you pick in a row, and I'll rebase it onto that.

The one setting it carries is the value colour, white by default, with magenta and blue as the two others that stay visible at brightness 1 in a dark room. If you'd rather it not be a setting, I can fix one colour and document it in the README.

ktomy:
You know, maybe we are not at the point where faces may become configurable, so maybe we can have per-face configuration, in your case there could be config for your face, something like "in-range color" "high color", "very high color", "low color", "very low color". Also think if you want the "simple" as the base or "big text"

ktomy:
Be aware that I've added another clock face meanwhile, so you'll have to adjust the face ID and selection

ktomy:
(ignore for this PR): another idea would be to add a selection for "enabled" faces. Now we have just a dropdown and in normal (not cycling) mode all clock faces are enabled, But as we'll have more and more faces, some of them highly dependent on environment (e.g. IOB/COB), for some people only a part of faces will be relevant. We already have a part of the infra for this in the cycling feature so I think now we are prepared for this feature... Maybe you could try to implement this as a different PR, not related to this one. Or I can implement this in a few weeks

nishanm:
Simple as the base: at night the smaller text is the point, and big text is what wakes people. Big text (dark) stays on my fork.

Per-face config is the right home for this, and I'll take the five band colours rather than one knob, because they add something: in-range magenta with the urgent tiers left red keeps the value calm all night and lets it light up only when something is actually wrong, which today's face can't do.

The one thing I'd hold is the default. All five should start white, so out of the box the face is one calm colour, which is what it exists for, and a band colour is something a user opts a tier into. If the defaults were the normal green/yellow/red, this would just be Simple with a coloured arrow and the sleep-through case would only exist for people who reconfigure five selects. The arrow keeps carrying the band either way, so nothing is hidden.

I'll build it that way, rebase for the Unicorn id (this becomes 7), and un-stack it from #190 so it can go in on its own. Enabled faces I'll take as a separate PR once the schedule is in.

nishanm:
Rebuilt as discussed: one commit on top of 0.31, face id 7 after Unicorn, no longer stacked on #190. The colour moved into a per-face block, `face_simple_dark`, with the five band colours, all defaulting to white; the fieldset sits under Device settings next to the data-old colour. Description updated. Builds for ulanzi.

nishanm:
Scratch my earlier comment on the day/night split. That's what #190's schedule does now (Big text in the morning, this face at brightness 1 in the evening), so this face does have a use.

The part that still stands is the settings. Five band colour selects make Device settings too heavy, and the schedule card will sit in the same section. I'd rather cut this to a single fixed value colour documented in the README, or show the block only when this face is selected. Which do you prefer?

ktomy:
> how the block only when this face is selected

This, I was thinking to have a block of active faces, a switch for cycling plus a dropdown or textbox for interval, then for each active face I would have a drawer with specific settings (for those which need settings), same as for the advanced settings

nishanm:
Makes sense. I'll build the Active faces block (enabled faces, cycling, a drawer per face) as its own PR first, then rebase this onto it with one number colour in the face's drawer; the arrow keeps the band, and stale uses the data-old colour like every other face.

nishanm:
Rebuilt as discussed: one number colour in a Simple (dark) drawer on the active faces block (via #178), the arrow keeps the range colour, stale uses the data-old colour, and urgent numbers turn red. Description updated.

ktomy:
Rebased onto current `main` (`d4ba07c`), including the rebased early-stale dependency from #178, and resolved conflicts.

Kept the daily schedule alongside the Simple (dark) settings drawer. Combined the configurable fresh-arrow color with main's stale-reading X. Compressed assets remain generated from source; raised the page budget to 42 KB for the added face controls (current output is 40,238 bytes).

nishanm:
One change on top of your rebase: the number color is now any color the firmware can draw except black, rather than white, magenta or blue.

Red, yellow and green are among them because this face holds the number in one fixed color and the reading's own color stays on the trend arrow, so they can't be read as a range; urgent readings still take the urgent color. The page still refuses a number color equal to the old-data color.

Eight text buttons don't fit a phone, so the picker is now labelled swatches. The old-data color picker already looked like this, so both share one helper.

Firmware and LittleFS builds pass, JavaScript syntax checks pass, and the page is 40,458 B against the 42,000 B budget. Checked against a mock clock: each color reaches the clock and survives a reload, gray is refused while old data is gray, and saves once old data is cyan. Hardware behavior has not been tested.

PR #192, opened 2026-09-16
Rebuild the settings page without Bootstrap or jQuery

nishanm:
The settings page is rebuilt as one self-contained page: about 26 KB gzipped instead of about 370 KB with Bootstrap, jQuery and the icon font. It's laid out for phones first and needs nothing from the internet, so setup works on the clock's own network. Every setting v0.31 saves is still saved, with the same validation.

Two things go against AGENTS.md, so I'd rather ask:
- The page lives in `web/src` and `node web/build.mjs` builds it into `data/index.html.gz` (Node built-ins only, no packages, output committed). If you'd rather have no build step, it can be one hand-written `data/index.html`, about 88 KB uncompressed.
- Controls are built in JS rather than in markup.

This PR updates AGENTS.md and CONTRIBUTING.md to match; say if you'd rather keep those yourself. Beyond v0.31: saving during first setup tells you to reconnect your phone to your WiFi, "Load from Nightscout" works with an API secret (the old page's hashing fails over http), there's Discard, and limits convert when you change units. The GitHub update check is kept and gives up after 5 s without internet.

`data_dev/` is no longer used; should it go too? #178, #187, #190 and #191 change the old page, and I'll port whichever merges first into this one.


nishanm:
One more commit: the custom no data timer moves from its own card on the Glucose tab into the color when data is old card on the Display tab, since it decides when a reading turns that color, and a change to it now marks the Display tab. The page's comments say color. Every PR built on this one is rebased onto it, and all of them now spell color the American way.

ktomy:
I am working on this, please don't commit

ktomy:
You know, the problem with this PR is that from now on it will be impossible to work on the UI without an AI agent. The changes are too big and too radical. I will accept the PR, but please, be aware that you will have to participate in the support activities as now a significant part of the whole thing if written by you

nishanm:
For sure. I’m hoping we get it to a point where it’s gold plated.

PR #193, opened 2026-09-16
Add a Time only face

nishanm:
For #132: a face that shows only the time, HH:MM:SS or H:MM AM/PM, so the clock can look like an ordinary clock.
- It switches to the Clock and value layout whenever the reading is urgent, and while an alarm sounds or is snoozed. With old or missing data only the time shows.
- To tick the seconds, a face can ask for a redraw every second (`ticksEverySecond()`); every other face keeps the per-minute refresh. The seconds pause while a data request is in flight — the per-second limitation you mentioned.


ktomy:
Rebased onto current `main` (`d4ba07c`) and resolved the generated `data/index.html.gz` conflict by keeping that file untracked and rebuilding it from `web/src`. Dropped the old settings-page predecessor commits already integrated into `main`; the Time only face source changes merged cleanly with active-face selection and scheduling.

PR #194, opened 2026-09-16
Show a cross instead of the trend arrow once the reading is old

nishanm:
Past the data-is-old threshold, Simple, Big text and Value and delta still drew the trend arrow for a reading the clock treats as too old. A cross in the data-is-old colour takes its place, at the same moment the value changes colour, including with the custom timer.
- A cross rather than a blank, because a blank slot already means a fresh reading with no trend.
- Graph and value and Clock and value already drop their one-column trend dots when old; there's no room for a mark there.


ktomy:
I don't fully agree with this approach as I want to see what was happening when the last value came, e.g. if the value was too high and the arrow was double-up, then my worry for the child is instantly higher than if I would see a cross. But you may have a point, so merging it with a small technical change

PR #196, opened 2026-09-16
Let a clock be given a name in the settings page

nishanm:
With several clocks in one home, every settings page is titled "Nightscout clock", so it's easy to change the wrong one. A pencil beside the heading opens a name field (suggests Kitchen, Bedroom, …; up to 32 characters). The name is saved as `clock_name` and becomes the page heading and tab title; empty keeps the default.

No firmware change: `/api/save` stores the key as sent, and the firmware's own saves update the file they read. The default is in `config_initial.json`.


ktomy:
Rebased onto current `main` (`d4ba07c`) and resolved conflicts. Combined the clock-name editing functions with main's documented header/status code, retained the current settings-page and scheduling implementation, and kept compressed assets generated from source rather than tracked in Git.

ktomy:
Hey @nishanm , is there a reason why you did not use the "Custom hostname" functionality for this? You could have removed the custom hostname setting in the advanced settings and put it up there with the edit functionality, so that it shows something like "Nightscout clock - kitchen.lan" in the top, with the pencil icon letting you edit the "kitchen.lan" part

nishanm:
I use the IP to reach the clocks — they're in the DHCP list and I have them bookmarked — so the hostname never gets typed here. The name is answering a different question: not "how do I reach this clock" but "which clock am I looking at" once the page is open.

There are also three practical snags with merging them. The firmware has no mDNS, so the hostname only becomes resolvable if the router happens to register DHCP names — showing `kitchen.lan` in the header would advertise an address that doesn't work on a lot of networks. `WiFi.setHostname()` runs once in `ServerManager::setup()` and `getHostname()` caches into `settings.hostname`, so a rename can't apply without a restart, which cuts against #198. And `getHostname()` also names the setup AP, so renaming the clock would rename the Wi-Fi network in the onboarding instructions.

Happy to make them agree rather than duplicate: if no name is set, the header could fall back to showing the hostname, so there's one identity on screen without the label having to be a valid hostname.

PR #197, opened 2026-09-16
Download and load the settings as a file

nishanm:
A Backup and restore card on the System tab: Download saves the clock's settings as a file, Load fills the form from one, and nothing is sent until Save.
- Loaded values are checked with the page's own option lists and `validateConfig`; a bad value keeps the clock's value and is listed. Keys the clock doesn't have are ignored.
- WiFi and network keys load only when ticked (another clock's static IP would cut this one off); the web login never loads.
- The file holds the WiFi and data source passwords in plain text, and the card says to keep it private.

No firmware change. Depends on #192; only the last commit is new.


ktomy:
Rebased onto current `main` (`d4ba07c`) and resolved conflicts in the settings cards, form API, and schema.

Preserved main's current UI and documentation while adding backup/restore. Updated file-import options to use `inactive_faces` and validate daily schedule entries. Kept compressed assets generated from source and allowed 42 KB for the expanded settings page.

Firmware and LittleFS builds, JavaScript syntax checks, and targeted backup-import checks for active faces, valid schedules, and malformed schedule entries pass. Hardware behavior has not been tested.

PR #198, opened 2026-09-16
Apply saved settings without restarting the clock

nishanm:
Saving settings no longer restarts the clock. The main loop loads the saved file and rebuilds what `setup()` derived from it: glucose ranges, face cycle, alarms, time zone and brightness. A snoozed alarm stays snoozed unless its own settings changed.
- WiFi, host name and the data source are still read only at startup, so the page restarts the clock when one of those changes (or in setup mode), and the button says "Save and restart" instead of "Save".
- `/api/save` only writes the file and sets a flag; the loop does the reload. Settings loads and saves, and the web server's login reads, share one lock.


ktomy:
Rebased onto current `main` (`d4ba07c`) and resolved the display-manager and settings-page conflicts.

Adapted live reload to `inactive_faces` and the daily schedule: schedule, timezone, and face-setting changes rebuild the schedule state, while unrelated saves preserve the current face. Combined the optional-restart save flow with the current UI/API structure and documentation. Kept compressed assets generated from source.

PR #199, opened 2026-09-16
Let each glucose range have its own colour

nishanm:
A colour per glucose range (`bg_color_urgent_low` … `bg_color_urgent_high`), defaulting to today's red/yellow/green/yellow/red, so nothing changes out of the box. Five settings because a colour-blind user can't tell red from green, and some families give a range its own meaning.
- Glucose tab: each range card has a swatch and a pencil that opens a named-colour dropdown.
- The page refuses gray and black (dark at brightness 1), an out-of-range colour equal to in range, urgent equal to its low/high neighbour, and any range equal to the old-data colour, so a reading can't look milder or stale than it is.
- Firmware: one `getBandColor()` used by the text faces, the graph dots and the Unicorn mane outside range (its in-range rainbow stays); `BG_COLOR_*` removed.

Two questions:
- Cyan, blue and magenta are also old-data options. Keep them for ranges (blocked only when they match), or drop them?
- The age bars and the trend dots keep their fixed green/yellow/red. Should they follow the range colours too, in a separate PR?


ktomy:
Rebased onto current `main` (`d4ba07c`) and resolved the generated web-asset conflicts. Removed the superseded settings-page predecessor commits and kept `data/index.html.gz` generated from `web/src`, preserving the per-range color controls alongside main's active-face selection, scheduling, and stale-reading X behavior.

PR #200, opened 2026-09-16
Add a Big text (dark) face

nishanm:
Big text (dark), face 8: Simple (dark)'s rules with Big text's large digits, for reading across a dark room. The calm number colour, red urgent numbers, the arrow in the range colour, and the data-old colour when stale all carry over.

Both dark faces draw through one `showDarkReading()` and share `face_simple_dark.value_color`, so the Clock faces card shows one drawer titled with whichever dark faces are active. No new config keys.

Depends on #191; only the last commit is new.


ktomy:
Rebased onto current `main` (`d4ba07c`), preserving the rebased early-stale and Simple (dark) dependencies from #178/#191. Resolved the generated `data/index.html.gz` conflict by keeping compressed assets generated from source. The shared dark-face settings retain main's daily scheduling and the stale-reading X behavior from the resolved dependency.

nishanm:
Rebased on the updated #178 and #191, so this face picks up the wider number color choice. Both dark faces still share the one color setting, and the drawer is titled with whichever of them are active.

Firmware and LittleFS builds pass, and the page is 40,559 B against the 42,000 B budget. Hardware behavior has not been tested.

PR #201, opened 2026-09-16
Let the Unicorn's mane move

nishanm:
The Unicorn by @JuanMiste (#181) is redrawn, and gets an optional moving mane. This changes the face's default look: a larger head with ears and a mane of six stepped colour bands that break into dim tips, magenta to gold in range and the low, high or urgent colour outside it (every other band darker). The reading and the age bars now end at the right edge, clear of the mane. The colours were chosen to stay distinct at brightness 1: the in-range colours and the gold horn keep their blue channel lit, where an orange or gold would show as the urgent red or the high yellow.
- Still stays the default. Moving (calm, normal or lively) is in the Unicorn drawer of the Clock faces card.
- In range the face's own rainbow rolls. Low and high keep their colour with a lighter stripe running through, with urgent red mixed in near the urgent limit, so green only ever means in range.
- Past an urgent limit, full-brightness dark stripes run through the mane (still lit at brightness 1); old data stops it in the old-data colour.
- Faces can return an animation step, and the manager redraws only the moving part. The motion colour, warning quarter and speed steps live in `BGDisplayFace` for later animated faces. `getGlucoseIntervals()` now returns a const reference.

**Optional, as its own commit so it can be dropped:** a Motion choice for the moving mane (`face_unicorn.flow`, default down): the colours flow down the bands, slide back toward the tips, or hold while a light runs along the bands. Without it the mane rolls down as described above.

**Also optional, its own commit:** an edge rule for the moving mane that replaces the urgent mix above. In the outer quarter of the in-range, low or high band every third band shows the colour of the band beyond that edge (green next to the in-range limit, red next to the urgent limit); the middle half shows one colour, and urgent and still manes are unchanged. It can be dropped like the Motion choice and the drifting tips.

Depends on #178; the last three commits are new.


nishanm:
Three commits added. Lively animations now step every 50 ms. The in-range mane is magenta, orchid, purple, violet, indigo and blue: after the LED gamma the old pink, coral and peach bands showed a red hue at brightness 2, next to the urgent color. The run flow's light takes the next band's color at irregular times instead of near-white, and the reading sits two columns from the tips again. Each is its own commit.

nishanm:
One more commit: the low and high flecks are now the band at 80% instead of near white, so a moving mane no longer carries white flecks beside the white head. Below an eighth of the brightness range they stay light, because a darker fleck lights the same LEDs as the band there and the motion would stop showing. #202 and #203 are rebased onto it, with their README rows saying darker stripe.

nishanm:
Id hold up on merging this one as i just cant get this one right at brightness 1.

nishanm:
One more commit, and the hold is lifted: the lowest brightness now looks right.

- **In range**, the run flow's bands hold their colors while single cells twinkle at random (a flash of the next band's color, then a darker shade), instead of lights running left to right.
- **At the lowest brightness** each LED channel is only on or off, so the in-range colors had collapsed to magenta and a solid blue block. There the mane is now magenta, blue and cyan from the top, two bands each, twinkling; two in five twinkles flash the glucose color (green, yellow or red). No cell goes dark and nothing turns white. Brightness 2 and up use the normal colors.

Low, high and urgent readings keep the behavior from the earlier commits. Checked in the emulator frame by frame against the approved design (440 frames, all identical; the previous head differs in every one), and both envs build.

ktomy:
Rebased onto current `main` (`d4ba07c`), preserving the rebased early-stale dependency from #178, and resolved conflicts.

Combined animation state with main's schedule state in the display manager and retained the Unicorn settings drawer alongside the daily schedule UI. Kept compressed assets generated from source and allowed 42 KB for the additional face controls. The full Unicorn animation commit sequence is preserved.

PR #202, opened 2026-09-16
Add a Dragon face

nishanm:
A Dragon face (id 7): a little dragon breathing fire at the reading, with a calm, normal or lively flame in its drawer.
- In range the fire is white, gold and peach, so it never reads as high or urgent (it shows as plain white at brightness 1). Outside range the flame uses the moving Unicorn mane's colours, including full-brightness dark stripes past urgent.
- The flame ends one dark column before the reading, so it never touches the digits.
- Old data puts the fire out and draws the dragon faded in the old-data colour, so it reads as stale even in blue or cyan.

Uses the step, quarter and motion-colour helpers from #201. Depends on #201; only the last commit is new.


nishanm:
Rebased onto #201's head. The in-range fire now uses the first four of #201's in-range colors, magenta to violet, instead of white, gold and peach, which also showed a red hue at brightness 2.

ktomy:
Rebased onto current `main` (`d4ba07c`), retaining the matching rebased early-stale and Unicorn animation dependencies. Resolved the settings-card conflict by keeping both the Dragon controls and daily schedule. Compressed web assets remain generated from source.

PR #203, opened 2026-09-16
Add a Race car face

nishanm:
A Race car face (id 7): a white car with speed lines in the glucose colour, a sliding road and spinning wheels, with a calm, normal or lively speed in its drawer.
- The speed lines use the motion colours from #201, so they match the moving Unicorn mane: a lighter stripe when low or high, dark stripes past urgent (still lit at brightness 1).
- The car, wheels and road are white, magenta, blue and cyan, never a glucose colour at any brightness. Old data stops the race in the old-data colour.

Depends on #201; only the last commit is new. #202 also takes face id 7, so whichever merges second moves up one.


nishanm:
Rebased onto #201's head; the Race car itself is unchanged.

ktomy:
Rebased onto current `main` (`d4ba07c`), retaining the matching rebased early-stale and Unicorn animation dependencies. Resolved the settings-card conflict by keeping both the Race car controls and daily schedule. Compressed web assets remain generated from source, and the later stale-car rendering fix is preserved.

PR #206, opened 2026-09-16
Show a live preview of the clock in the settings page

nishanm:
**Draft, and a question before any review.** The trade-off first: this adds a live clock preview to the settings page, so every display-code change (every face PR) would then need Emscripten to rebuild `data/clockemu.js.gz`, and the filesystem grows by about 64 kB (preview 53.6 kB, page 26.2 → 37.0 kB, both under the budgets in `web/build.mjs`). If that toolchain cost isn't worth it to you, I'm happy to close this.

What it does: the firmware's own face and display code runs as WebAssembly in the browser, drawing the clock with the page's unsaved settings, the three buttons and a test reading. Nothing loads from the internet, it only runs while visible, and the page works as before if the file is missing.

The second commit is the page design built around the preview:
- The preview sits in its own column beside the settings on a wide screen, and above them on a phone, where a bottom tab bar moves between tabs.
- Each face tile carries a picture drawn by the firmware and a one-line description; tapping a tile shows it in the preview.
- The preview says how many LEDs the face lights and warns when some go dark, or lose a colour channel, at the chosen brightness, so a colour that vanishes at level 1 is seen before it is saved.
- Hold ◀ / Hold ▶ (long press) and Select ×2 (double click) work as on the clock.
- Try a reading: a strip in the range colours under the glucose slider, a trend picker, a reading that is fresh, one missed, past the no-data time or absent, a clock time that can be set and run faster, a room light slider for the automatic brightness modes, and Animate and Sound.
- Each alarm's sound can be played here, through the firmware's melody code, next to Try on clock.
- The Time card offers this browser's time zone whenever the clock is in another one.
- A setting missing from the clock's file takes the factory value (`config_initial.json`), as the firmware does.

The source and build script are in `web/emulator/`; `web/emulator/build.sh` then `node web/build.mjs` rebuild it, and the same checkout rebuilds to the same bytes. It's a preview, not a substitute for testing on the clock.

Depends on #192; the last two commits are new.


nishanm:
One commit added: the "LEDs lose a color channel" warning under the preview now shows the worst of the last 3 s. On an animated face it used to blink on and off from frame to frame, which moved the page. It starts afresh when the face, the brightness or any setting changes. The rebuilt page differs from the previous one only in those lines.

ktomy:
Rebased onto current `main` (`d4ba07c`) and resolved the settings-page, build-script, and documentation conflicts.

- Combined preview thumbnails and controls with active-face selection, the default-face selector, and the daily schedule card.
- Preserved main's documented API/form helpers and generated web assets; allowed 52 KB for the expanded page (50,114 bytes).
- Rebuilt the committed browser emulator with Emscripten 6.0.9 against current firmware, including active faces, daily scheduling, and stale-reading rendering.

Firmware and LittleFS builds and JavaScript syntax checks pass. Headless Chrome checks confirm the mobile preview loads, face toggles work, and the rebuilt emulator applies a scheduled face and skips inactive faces. Hardware behavior has not been tested.

PR #208, opened 2026-09-19
Add middle-button brightness control

0m4r:
Cycle persisted manual brightness from 5% through 100% with compact icon feedback. Pause automatic face rotation while the value is shown and restore the previous automatic brightness mode with a long middle-button press.

Adds middle-button brightness control with persistent manual brightness settings and automatic brightness restoration.

- Short middle-button presses cycle through:
  - 5%
  - 20%
  - 40%
  - 60%
  - 80%
  - 100%
  - Then back to 5% 
- Selected manual brightness is persisted across restarts.
- If automatic brightness was active, the previous automatic mode is remembered.
- Long middle-button presses restore the previous automatic brightness mode.
- Displays a compact brightness icon with the selected percentage.
- Displays the brightness icon with AUTO after automatic mode is restored.
- Brightness feedback remains visible for 2 seconds and fades out over 0.5 seconds.
- Automatic clock-face rotation and regular face redraws pause while brightness feedback is displayed.
- Rotation resumes with its interval timer reset afterward.
- Updated the `README` with the new controls and behavior.

**Validation**
- Firmware builds successfully with `bash scripts/build.sh`.
- `README` and modified source changes pass whitespace validation.
- No LittleFS/filesystem changes are required for this feature; firmware-only upload preserves saved device settings.


ktomy:
Rebased onto current `main` (`d4ba07c`) and resolved conflicts in the display manager and README.

Kept main's active-face list and daily schedule state alongside the brightness overlay. Scheduled/cycling updates resume after the overlay clears. Preserved the stale-reading X documentation and the new middle-button instructions.

src/DisplayManager.cpp
```diff
@@ -50,22 +50,22 @@ void setMatrixLayout(int layout) {
     delete matrix;  // Free memory from the current matrix object
     DEBUG_PRINTF("Set matrix layout to %i", layout);
     switch (layout) {
-        case 0: // Ulanzi
+        case 0:  // Ulanzi
```
ktomy, line 53: Please remove these (whitespaces) changes as they do not add value

ktomy:
Hi, 

Thank you for your contribution!

Do I understand correctly that multiple presses change brightness multiple times? How do we then deal with the "double-click" functionality which turns off the display?
Also you made changes to the changelog, please remove them as I will be changing the versions and not sure in which exact version this feature will be released (actually I am not sure whether to increase the major version number due to the massive web UI change).
Also I would incline towards the full cycle, I mean presses not only change from 1 to 10 but also include auto modes. Also change has to be persisted in config
Also please remove the changes to the whitespaces for the regions which you did not really touch.

0m4r:
hi @ktomy thanks for the feedback.

The middle button preserves the old functionality of "double click" to turn on and off the screen (tested on physical device).
The brightness changes on "single click" and cycles to the default values (0% does not seem to make sense so I went for a 5% min value). 
A middle button "long press" would reset to auto-mode or default brightness as saved in the setting screen.

I will take care to remove the white spaces and the release note.
I will need to look into "_Also change has to be persisted in config_", as this would imply to have the brightness adjustments to match the values in the setting page and not the once I have used. But I have to say this would make sense.

0m4r:
hi @ktomy I have update the code and, accordingly, the PR description.
Let me know what you think about it!

